Issue #49, opened 2015-01-06
Unable to install model fragments w/ 0.1.5 and ember-data beta.14.1


thoov:
If I create a blank project with ember-cli v0.1.5 and update ember-data to beta.14.1 the app will load up just fine. However, after running:

```
ember install:addon ember-data-model-fragments
```

I get this error in the browser:

```
Uncaught ReferenceError: define is not defined
```

Not sure if I'm the only one getting this or if this is happening to others?

slindberg:
Can you give more detailed steps? It appears to work for me™

```sh
$ npm cache clear
$ bower cache clean
$ ember new app && cd app
version: 0.1.5
installing
  create .bowerrc
  create .editorconfig
  create .ember-cli
  create .jshintrc
  create .travis.yml
  create Brocfile.js
  create README.md
  create app/app.js
  create app/components/.gitkeep
  create app/controllers/.gitkeep
  create app/helpers/.gitkeep
  create app/index.html
  create app/models/.gitkeep
  create app/router.js
  create app/routes/.gitkeep
  create app/styles/app.css
  create app/templates/application.hbs
  create app/templates/components/.gitkeep
  create app/views/.gitkeep
  create bower.json
  create config/environment.js
  create .gitignore
  create package.json
  create public/.gitkeep
  create public/crossdomain.xml
  create public/robots.txt
  create testem.json
  create tests/.jshintrc
  create tests/helpers/resolver.js
  create tests/helpers/start-app.js
  create tests/index.html
  create tests/test-helper.js
  create tests/unit/.gitkeep
  create vendor/.gitkeep
Installed packages for tooling via npm.
Installed browser packages via Bower.
Successfully initialized git.
```

then

```sh
$ ember install:addon ember-data-model-fragments
version: 0.1.5
Installed packages for tooling via npm.
installing
The globPattern "undefined" did not match any files, so no file updates will be made.
Installing browser packages via Bower...
  cached git://github.com/components/ember-data.git#1.0.0-beta.12
  cached git://github.com/components/ember-data.git#1.0.0-beta.12
  new version for git://github.com/components/ember-data.git#~1.0.0-beta.7
  resolved git://github.com/components/ember-data.git#1.0.0-beta.14.1
Installed browser packages via Bower.
```

knownasilya:
What's `The globPattern "undefined" did not match any files, so no file updates will be made.`?

thoov:
[...]

PR #52, opened 2015-01-06
Added 'defaultBlueprint' override to fix lookup bug

slindberg:
Refs #49


slindberg:
See https://github.com/ember-cli/ember-cli/issues/2915

Issue #48, opened 2015-01-06
Error on save() when observing hasManyFragments property


dwickern:
Live example (hit save): http://jsbin.com/yibuniqami/1/edit?js,console,output

This happens when observing a hasManyFragments array, e.g.

`whoops: function() {}.observes('hasManyFragmentProperty.[]')`

My REST API responds with the server's version of the model. The client chokes with this error when it tries to update its version
```
TypeError: Cannot read property 'destroy' of undefined
    at ContainerView.extend.arrayWillChange (http://builds.emberjs.com/tags/v1.9.0/ember.js:40448:20)
    at applyStr (http://builds.emberjs.com/tags/v1.9.0/ember.js:19670:29)
    at sendEvent (http://builds.emberjs.com/tags/v1.9.0/ember.js:14106:13)
    at __exports__.default.Mixin.create.arrayContentWillChange (http://builds.emberjs.com/tags/v1.9.0/ember.js:30434:9)
    at EmberObject.extend.arrangedContentArrayWillChange (http://builds.emberjs.com/tags/v1.9.0/ember.js:34301:14)
    at applyStr (http://builds.emberjs.com/tags/v1.9.0/ember.js:19670:29)
    at sendEvent (http://builds.emberjs.com/tags/v1.9.0/ember.js:14106:13)
    at Array.__exports__.default.Mixin.create.arrayContentWillChange (http://builds.emberjs.com/tags/v1.9.0/ember.js:30434:9)
    at Array.Mixin.create.replace (http://builds.emberjs.com/tags/v1.9.0/ember.js:35815:14)
    at StatefulArray.extend.replaceContent (https://rawgit.com/lytics/ember-data.model-fragments/master/dist/ember-data.model-fragments.js:312:37)
```

If I change from observing `[]` to `@each` I get this error instead:
```
RangeError: Maximum call stack size exceeded
    at GETTER_FUNCTION [as _parentView] (http://builds.emberjs.com/tags/v1.9.0/ember.js:17067:38)
    at null.<anonymous> (http://builds.emberjs.com/tags/v1.9.0/ember.js:41363:26)
    at Descriptor.ComputedPropertyPrototype.get (http://builds.emberjs.com/tags/v1.9.0/ember.js:12122:25)
    at get (http://builds.emberjs.com/tags/v1.9.0/ember.js:17570:21)
    at CoreView.extend.destroy (http://builds.emberjs.com/tags/v1.9.0/ember.js:43560:36)
[...]

PR #54, opened 2015-01-07
Fix issue with array observer infinite recursion on adapter update

slindberg:
Refs #48


slindberg:
@dwickern This fixes the [example jsbin](http://jsbin.com/yibuniqami/1/edit?js,console,output) you created, but would you mind testing this in your app in case there are other scenarios I'm not accounting for?

dwickern:
It's working now. Thanks!

slindberg:
Great, thanks for reporting this and all your help. Here's a panda for you :panda_face:

Issue #56, opened 2015-01-08
Multiple rollbacks don't work starting with 0.2.4


workmanw:
See this JSBIN: http://emberjs.jsbin.com/yemimobire/5/edit

This is working with 0.2.3, but broken with 0.2.4 and 0.2.6 (didn't bother checking 0.2.5).

=====

**Edit: Sorry for the poor description.**

The first time you call rollback on a model, it works correctly (restoring the original value and setting isDirty back to false), the second time you call rollback on that same model it sets isDirty back to false, but does no restore the value.

slindberg:
Thanks for testing this in your app and reporting, it looks like this is happening [with `hasManyFragment` properties as well](http://emberjs.jsbin.com/xedunuwaza/1/edit?html,js,output). Sorry for all the issues with this upgrade, the shift away from observing the `data` property had a lot of unseen consequences. I'll take a look at this as soon as I can.

/cc @igorT

workmanw:
Hey no worries at all! Happy to. It just so happened we had an integration test that caught this one.

slindberg:
@workmanw would you mind testing the fix in your app using the `tmp-dist` branch? you can use [this file](https://rawgit.com/lytics/ember-data.model-fragments/tmp-dist/dist/ember-data.model-fragments.js) for jsbins or put this in your `bower.json`:

```json
{
  "dependencies": {
    "ember-data.model-fragments": "lytics/ember-data.model-fragments#tmp-dist",
    ...
  }
}
```

workmanw:
:smile: Confirmed fix!

Just to note, I first upgrade to 0.2.6 to confirm that our tests would still fail (they did). Then I upgraded to your #tmp-dist branch and our tests passed. Thanks!!!

slindberg:
Fixed in 0.2.7

workmanw:
Thanks again!

PR #57, opened 2015-01-14
Fix issue with multiple rollbacks

slindberg:
Refs #56

### DIFF
--- a/packages/model-fragments/lib/fragments/ext.js
+++ b/packages/model-fragments/lib/fragments/ext.js
@@ -172,7 +172,6 @@ Model.reopen({
       if (this._fragments[key]) {
         this._fragments[key].rollback();
       }
-      this._fragments[key] = null;
     }
   },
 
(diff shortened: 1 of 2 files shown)


packages/model-fragments/lib/fragments/ext.js
```diff
@@ -172,7 +172,6 @@ Model.reopen({
       if (this._fragments[key]) {
         this._fragments[key].rollback();
       }
-      this._fragments[key] = null;
```
slindberg: @igorT I'm not sure anymore [why this was added originally](https://github.com/lytics/ember-data.model-fragments/pull/34/files#diff-b8af553d86227a5fda84f3e123c722e3R972). It doesn't seem to affect any other tests, would you mind [testing it in your app](https://rawgit.com/lytics/ember-data.model-fragments/multiple-rollback-tmp-dist/dist/ember-data.model-fragments.js)?

PR #60, opened 2015-02-02
Fix infinite loops when reloading observed fragments

igorT:
Currently in certain annoying cases in Ember, including aliases or proxies that are actively observed,
CPs are consumed as soon as they are changed. This used to cause an infinite loop. For example:

If you had
```
  App.User = DS.Model.extend({
    name: hasOneFragment('name')
  })

  App.Name = DS.ModelFragment.extend({
    firstName: DS.attr()
  })
```

1. We would push new data for a user
2. We would notify a change for the user.name which cleared the fragments cache
3. An observer would try to get `user.name.firstName` synchronously
4. As we had new data, we would try to setupData on the fragment
5. In the middle of fragment data setup we would notify a change to `firstName`
6. The observer would fire immediately and try to get `user.name.firstName` even though
  we weren't done with the setup
7. Goto 4

We now cache the fragment early enough to prevent this infinite loop

### DIFF
--- a/packages/model-fragments/lib/fragments/attributes.js
+++ b/packages/model-fragments/lib/fragments/attributes.js
@@ -79,8 +79,10 @@ function hasOneFragment(declaredTypeName, options) {
     // Else initialize the fragment
     } else if (data && data !== fragment) {
       fragment || (fragment = setOwner(store.buildFragment(actualTypeName)));
-      fragment.setupData(data);
+      //Make sure to first cache the fragment before calling setupData, so if setupData causes this CP to be accessed
+      //again we have it cached already
       this._data[key] = fragment;
+      fragment.setupData(data);
     } else {
       // Handle the adapter setting the fragment to null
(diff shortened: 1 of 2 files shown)


slindberg:
Looks like this was originally avoided by using that [`Ember.beforeObserver`](https://github.com/lytics/ember-data.model-fragments/blob/6a5eef486d49c8b97f4a891296f9303203dac61c/packages/model-fragments/lib/fragments/ext.js#L80-L102) on the `data` attribute, which always made me queasy. Thanks for your help sorting out these edge cases :+1:

I'll do a bugfix release immediately unless you're aware of any other issues with the current build.

igorT:
Haven't noticed any other issues at the moment. App test suite passes with 0.2.7 with the above fix

slindberg:
In [v0.2.8](https://github.com/lytics/ember-data.model-fragments/tree/v0.2.8)

PR #66, opened 2015-02-18
Snapshot update (beta.15 compatibility)

slindberg:
The change to passing snapshots to serializers instead of records has interesting implications for model fragments. At their core, fragments are simply embedded JSON attributes, so from the perspective of a snapshot, they should be considered attributes and _not_ relationships. The `hasOneFragment` and `hasManyFragments` properties are already marked as attributes, but when creating a new snapshot, the fragment model/array instance is returned when calling `snapshot#attr` – not exactly a snapshot of the attributes.

Therefore fragments need to be serialized in the snapshot _before_ `DS.Model#serialize` is invoked, instead of afterwards when transforms are applied to attributes. This unfortunately means relying on even more private APIs, in this case `DS.Model#_createSnapshot` (since the Snapshot constructor is not easy to override).

@wecc I don't know how familiar you are with this project, but if you could look over where I'm hooking into snapshots, I'd love to get your feedback!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -21,12 +21,11 @@
     "test.html.mustache"
   ],
   "dependencies": {
-    "ember": "^1.4.0",
-    "ember-data": "^1.0.0-beta.14"
+    "ember-data": "^1.0.0-beta.15"
   },
   "devDependencies": {
     "qunit": "^1.14.0",
     "jquery": "^2.0.0",
-    "ember": "^1.4.0"
+    "ember": "^1.10.0"
   }
 }
--- a/packages/model-fragments/lib/core-model.js
+++ b/packages/model-fragments/lib/core-model.js
@@ -13,6 +13,7 @@ Model.proto();
 var protoProps = [
   '_setup',
   '_unhandledEvent',
+  '_createSnapshot',
   'send',
   'transitionTo',
   'isEmpty',
--- a/packages/model-fragments/lib/fragments/array/fragment.js
+++ b/packages/model-fragments/lib/fragments/array/fragment.js
@@ -79,6 +79,17 @@ var FragmentArray = StatefulArray.extend({
   },
 
   /**
+    @method _createSnapshot
+    @private
+  */
+  _createSnapshot: function() {
+    // Snapshot each fragment
+    return map(this, function(fragment) {
+      return fragment._createSnapshot();
+    });
+  },
+
+  /**
     @method adapterDidCommit
   */
   adapterDidCommit: function() {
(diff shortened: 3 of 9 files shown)


slindberg:
For reference: emberjs/data#2623

igorT:
Hm, I am not sure you really need to do this, but seems hard to think about. Just creating a snapshot should be fully backwards compatible though, if it's not we screwed something up. Do you have time tomorrow for a quick chat?

slindberg:
> Just creating a snapshot should be fully backwards compatible though

Yes, [cherry picking the `_createSnapshot` method](https://github.com/lytics/ember-data.model-fragments/pull/66/files#diff-2ac32cb2148644650b2a86804dd5ef27R16) from `DS.Model` is enough to get all tests passing, which is great.

I'm mainly trying to keep model fragments inline with the idea behind snapshots (although perhaps I've misinterpreted?). To me, that means that if a snapshot instance hangs around, attributes on the snapshot – including fragments – don't change. Although the more I think about it, the more it seems like maybe calling `snapshot.attr('fragmentProperty')` should return a snapshot instance(s), and not a pre-serialized fragment. Will think on this.

I'm available 9-5 pacific time, hit me up on IRC.

slindberg:
Thinking about this more, I realize that _serializing_ the fragment in the snapshot can result in confusion when fragments are accessed in the serializer, since attribute names can be munged by the serializer. I think the right approach is for `snapshot.attr('someFragment')` to return a snapshot of the fragment, or an array of snapshots.

I've rebased this PR and updated the `tmp-dist` branch.

workmanw:
@slindberg This gets a :+1: from me. Having `snapshot.attr('someFragment')` return a snapshot not only feels consistent, but seems better in the way of future proofing. It also means that my custom serializer doesn't have to support two code paths (snapshot and fragmentRecords).

The `tmp-dist` branch seems to work great in our app w/ `ember-data-1.0.0-beta.15`. All tests pass. Thanks a lot!

slindberg:
@workmanw awesome, thanks for testing with your app, super appreciate it. This is working for my app too, so I'm going to merge this in today or tomorrow and create a release unless @igorT or @wecc have any objections.

blimmer:
I'm testing the upgrade to ED 1.0-0-beta.15 on my app, so I'll confirm this PR locally and update shortly.

blimmer:
This PR does not work for me. I get the following error when trying to save the model:
```
'undefined' is not a function (evaluating 'record._createSnapshot()') (/Users/blimmer/code/Ibotta-Web/public-test/javascripts/vendor.js:71931)
```

The `tmp-dist` branch, on the other hand, works great!

slindberg:
Thanks @blimmer :+1:

I'll try to get this out sometime this afternoon.

PR #69, opened 2015-02-19
Distribution files have wrong version numbers

jreading:
Unless I'm woefully mistaken, the files in the dist dir have the wrong version numbers. Running `grunt dist` only changes version numbers.


slindberg:
Yeah, I goofed on the last release. Can't republish the NPM package though, so we're stuck with the mismatch until the next release. Sorry for any confusion it caused!

jreading:
Might not be a bad thing for bower though to just fix master and retag v0.2.8. Thanks anyway!

slindberg:
You've got a good point, and I considered doing it when I found out, but finally decided I'd rather avoid even more confusion if someone sees the mismatch, then compares bower and NPM versions to find that _they_ aren't even the same :astonished:

PR #85, opened 2015-04-30
Add .bowerrc so bower files are placed correctly

knownasilya:
Fixes the test runner

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory": "bower_components"
+}


slindberg:
Is this necessary if you are overriding the bower directory in your app? I'm all for completeness but I don't understand what problem this is actually solving.

knownasilya:
Bower by default places the files in `app/bower_directory`, and your test html is looking for it in `bower_directory` (sans `app/`).

slindberg:
I'm afraid I'm still at a loss :confused: 

> Bower by default places the files in app/bower_directory

How does this happen? My understanding is that [bower uses `bower_directory` by default](http://bower.io/docs/config/#directory).

For a sanity check, I did a fresh clone and followed the [testing instructions](https://github.com/lytics/ember-data.model-fragments#testing), and encountered no errors. Since Travis builds are passing as well, I'm inclined to believe that the bower components are being correctly installed. Am I missing something?

PR #90, opened 2015-05-14
Update to use new computed property syntax

jakesjews:
We were getting a lot of deprecation errors when we upgraded our application to Ember 1.12. This converts the hasOneFragment and hasManyFragments properties to use the new computed property syntax.
I tried using the [ember-new-computed polyfill](https://github.com/rwjblue/ember-new-computed) but I wasn't able to get it to import into the project.


slindberg:
Thanks for the work! I understand how annoying deprecation warnings are, especially when they come from a dependency. It is my understanding that the new CP syntax is only valid as of 1.12.0 though, so I'm a bit concerned about backwards compatibility. I'm not sure what the best way forward is... perhaps some Ember feature detection?

jakesjews:
It would be cool to use the new-computed property polyfill but I had a hard time getting it to install. I assume its because ember-cli isn't being used as the build system. I could probably copy in the code for ember-new-computed by hand if you don't mind a little code bloat.

slindberg:
You're right, it's because the source is pre-built, much like ED is. I hesitate to make this addon pull source into the dependency tree, since build times are already too long for most reasonably sized apps IMO. Adding babel (and the annotations babel plugin) as a dependency would just increase that build time.

The solution here I think it to put in some checks to use the new CP syntax for apps using v1.12.0, and continue to use the old syntax for pre-1.12 versions.

jakesjews:
Sounds like a plan. I'll see what I can come up with.

aars:
+1

rwjblue:
Take a look at https://github.com/emberjs/ember-inspector/pull/352. The same technique should work here...

slindberg:
:+1: to the approach used by ember-inspector. @jakesjews do you want to tackle this?

jakesjews:
Yeah I'd like to. I'm pretty busy today but I should be able to get it tomorrow.

GeoffreyBooth:
+1

jakesjews:
Just got it in. Tests are passing locally so hopefully it'll be ok in Travis. Is the folder structure ok?

packages/model-fragments/lib/fragments/attributes.js
```diff
@@ -52,20 +53,19 @@ function hasOneFragment(declaredTypeName, options) {
...
-    var data = this._data[key] || getDefaultValue(this, options, 'object');
-    var fragment = this._fragments[key];
-    var actualTypeName = getActualFragmentType(declaredTypeName, options, data);
+  function setOwner(fragment, record, key) {
+    Ember.assert("Fragments can only belong to one owner, try copying instead", !get(fragment, '_owner') || get(fragment, '_owner') === record);
+    return fragment.setProperties({
+      _owner : record,
+      _name  : key
+    });
+  }
```
slindberg: Since this function is no longer using `record` and `key` as closures, how about pulling it out to the top level module scope and renaming it `setFragmentOwner`?

slindberg:
Awesome, thanks!

I would like to avoid the potential scenario where someone sees that they can import the computed shim into their app, only to have it break when it gets removed in a non-minor version bump. How about putting it under a new directory `/packages/model-fragments/lib/util`? I also don't see a reason to keep the `supportsSetterGetter` module separate, so can you just include it at the top of the computed shim module?

jakesjews:
Alrighty, I pulled setOwner out to the top level scope and moved the computed polyfill into a util folder

slindberg:
:beers:

PR #102, opened 2015-06-11
[PENDING] Update for renamed DS.Snapshot -> DS.AdapterSnapshot

HeroicEric:
fixes https://github.com/lytics/ember-data.model-fragments/issues/101


HeroicEric:
decision was made to not rename in Ember data

PR #109, opened 2015-07-12
Update to Ember Data v1.13

### LINKED ISSUE #99: Uncaught TypeError: Cannot read property '_unhandledEvent' of undefined with Ember Data 1.0.0-beta.19
When upgrading Ember Data to 1.0.0-beta.19, this error appears (from `ember-data.model-fragments.js:98`):
```
Uncaught TypeError: Cannot read property '_unhandledEvent' of undefined
```
And Ember Data Model Fragments fails to load, which causes other errors to be thrown. 

See jsbin: http://emberjs.jsbin.com/hafaso/1/edit?html,js,console. Change Ember Data from beta.19 to beta.18 (HTML tab line 12) and the error goes away.

slindberg:
Closes #99.


packages/model-fragments/lib/fragments/ext.js
```diff
@@ -152,83 +100,141 @@ Model.reopen({
...
     return diffData;
   },
+});
 
-  /**
-    If the model `isDirty` this function will discard any unsaved
-    changes, recursively doing the same for all fragment properties.
+// Replace a method on an object with a new one that calls the original and then
+// invokes a function with the result
+function decorateMethod(obj, name, fn) {
```
slindberg: @igorT the diff is a little difficult to read, but [these frankenstein method decorators](https://github.com/lytics/ember-data.model-fragments/blob/internal-model-compat/packages/model-fragments/lib/fragments/ext.js#L117-L195) are one of the terrible things I mentioned.

I tried pretty hard to stay away from touching `InternalModel`'s prototype, but was unable to accomplish what these three methods do without doing so:

1. `createSnapshot` - since fragment attributes are not relationships, they are accessed through `Snapshot#attr`, which means that by the time the user sees a record snapshot, its fragment attributes must already be snapshotted. The alternative was doing something similar to `Snapshot`'s prototype...
2. `rollbackAttributes` - This could potentially have stayed in the `Model.reopen()`, but then it runs the risk of getting out of sync if `InternalModel#rollbackAttributes` is ever [called directly](https://github.com/emberjs/data/blob/23a5b13a9bccd3f26515f4992dd59960d6649df0/packages/ember-data/lib/system/model/states.js#L276).
3. `adapterDidCommit` - It doesn't look like there's any other hook where updating fragment states can happen. This one is particularly nasty because of the need to use the original method's `data` argument.

packages/model-fragments/lib/fragments/ext.js
```diff
@@ -152,83 +100,141 @@ Model.reopen({
...
-
-    // Don't reset if the record is new, otherwise it will enter the 'deleted' state
-    // NOTE: This case almost never happens with attributes because their initial value
-    // is always undefined, which is *usually* not what attributes get 'reset' to
-    if (!get(this, 'isNew')) {
-      this.send('propertyWasReset', key);
+  */
+  transformFor: function(attributeType) {
+    if (attributeType.indexOf('-mf-') === 0) {
+      return getFragmentTransform(this.container, this.store, attributeType);
```
slindberg: @igorT This is the other terrible thing I've done. This actually wasn't strictly necessary for compatibility with v1.13; it's just [something I've wanted to do for a while](https://github.com/lytics/ember-data.model-fragments/issues/12), and [keeps coming up](https://github.com/lytics/ember-data.model-fragments/issues/95). The gist is that when fragment attribute CPs access the data in `._data`, it should already be deserialized, which gives people an opportunity to override fragment transforms if they want/need.

Since the transform needs to know things about the specific fragment attribute (attribute type, fragment type, and polymorphic type key), and passing enough information to `Transform#deserialize` in [`applyTransforms`](https://github.com/emberjs/data/blob/v1.13.5/packages/ember-data/lib/serializers/json-serializer.js#L195) doesn't seem feasible, the only solution I could come up with is to create a new transform class. At first I balked at the thought, but then I came around, since there's very likely only a handful of transforms that are created in a given app.

The bad:

- This overrides a private method on `JSONSerializer`, which is never a good idea
- Relying on `JSONSerializer` breaks apps that have completely custom serializers (which I don't think is common, but still)
- Encoding type information as a string in the attribute type meta smells thoroughly. I'm not even sure if the `$` delimiter is intended to be a valid character in container keys

packages/model-fragments/lib/fragments/model.js
```diff
@@ -241,5 +135,29 @@ function getActualFragmentType(declaredType, options, data) {
...
+  var internalModel = record._internalModel;
+
+  // Ensure the internal model has a fragments hash, since we can't override the
+  // constructor function anymore
+  if (!internalModel._fragments) {
+    internalModel._fragments = create(null);
+  }
+
+  return internalModel;
+}
```
slindberg: @igorT Overall, I was unhappy with how heavily this relies on `_internalModel`, because it seemed like a step backward. In retrospect, I think it's more of a side-step, since MF has always made use of private APIs like `._data`. I'd love to see it move away from touching model internals, but there are certain things that are unavoidable, i.e. storing state on the model instance.

This method is another `InternalModel` workaround, since there's no longer an `init` hook I can take advantage of to ensure the `._fragments` hash exists.

slindberg: On a related note, methods like `Ember.create` are deprecated, and ED is using its own polyfills to avoid those warnings for v1.13, but they aren't exposed globally. Just copy-pasting them into this project seems wasteful, but I don't really see another way to prevent deprecation warnings – do you have any suggestions?

devinus:
This is too much awesome at once.

devinus:
@slindberg I'm getting this while testing:

```
Error while processing route: campaigns.index Cannot read property 'hasOwnProperty' of null TypeError: Cannot read property 'hasOwnProperty' of null
    at applyTransform (http://localhost:4200/assets/vendor.js:75973:20)
    at http://localhost:4200/assets/vendor.js:86960:20
    at Map.forEach.cb (http://localhost:4200/assets/vendor.js:26547:11)
    at OrderedSet.forEach (http://localhost:4200/assets/vendor.js:26330:11)
    at Map.forEach (http://localhost:4200/assets/vendor.js:26551:18)
    at Function.ember$data$lib$system$model$model$$default.reopenClass.eachTransformedAttribute (http://localhost:4200/assets/vendor.js:86959:84)
    at ember$data$lib$system$serializer$$default.extend.applyTransforms (http://localhost:4200/assets/vendor.js:75972:19)
    at ember$data$lib$serializers$json$serializer$$default.extend.normalize (http://localhost:4200/assets/vendor.js:78972:14)
    at superWrapper [as normalize] (http://localhost:4200/assets/vendor.js:31286:20)
    at deserializeFragment [as deserialize] (http://localhost:4200/assets/vendor.js:90572:27)
```

devinus:
Seems to be blowing up here: https://github.com/emberjs/data/blob/2385a7c2e64467a1823b91795a07bf68d3940b4c/packages/ember-data/lib/serializers/json-serializer.js#L191

What would cause data to be null?

devinus:
Apprently a fragment cannot come back from the server as `null`?

devinus:
`hasOneFragment` can't handle the fragment coming back as `null`. I have to send it as `{}`

devinus:
Also, all the `isDirty` references should be changed to `hasDirtyAttributes`.

devinus:
Also, getting this now:

`Warning: ignoring input sourcemap for bower_components/ember-data.model-fragments/dist/ember-data.model-fragments.js because ENOENT: no such file or directory, open '/Users/devin/Projects/rebelmail-dashboard/tmp/tree_merger-tmp_dest_dir-YNIRQpli.tmp/bower_components/ember-data.model-fragments/dist/ember-data.model-fragments.map'`

packages/model-fragments/lib/fragments/model.js
devinus: Regarding this, the pattern I see emerging is people doing:

```
const keys = Object.keys || Ember.keys;
const create = Object.create || Ember.create;
```

That way, only terrible browsers get the deprecation warning. I'm not sure what the plan is regarding sharing these shims among packages. I don't know why Ember doesn't want to expose the shim themselves since it has to have the shim anyway.

slindberg:
> hasOneFragment can't handle the fragment coming back as null. I have to send it as {}

Working on a fix now, thanks for testing it out!

> Also, all the isDirty references should be changed to hasDirtyAttributes.

Are you referring to the references to `isDirty` in tests, or are you getting a deprecation warning in your code? [They should have all been updated](https://github.com/lytics/ember-data.model-fragments/commit/d8d76e5a6fd765d089be426960c8b57421b0b2ec).

As for the source map, I omitted that from the dist build. I'll add it back.

devinus:
@slindberg the code itself still has a reference:

https://github.com/lytics/ember-data.model-fragments/blob/tmp-dist/packages/model-fragments/lib/fragments/array/stateful.js#L183

slindberg:
> @slindberg the code itself still has a reference:

Well how about that.

devinus:
Also getting this:

`DEPRECATION: `lookup` was called on a Registry. The `initializer` API no longer receives a container, and you should use an `instanceInitializer` to look up objects from the container. See http://emberjs.com/guides/deprecations#toc_deprecate-access-to-instances-in-initializers for more details.` from `getFragmentTransform`

devinus:
I have a few `hasManyFragments` with `defaultValue: Ember.A()` but they're being defaulted to `null` right now.

packages/model-fragments/lib/fragments/model.js
slindberg: @devinus makes sense, I had just assumed that the ED object polyfills were full implementations.

slindberg:
@devinus I think I fixed all the deprecation warnings and pushed to the `tmp-dist` branch, but I don't know what's going on with the default array values. Can you tinker with [the tests](https://github.com/lytics/ember-data.model-fragments/blob/internal-model-compat/packages/model-fragments/tests/unit/fragments/has_many_test.js#L234-L276) to get one to fail?

Thanks again for all your help!

devinus:
@slindberg Only more insight I can give right now is I have to again use `{ defaultValue: {} }` for the fragment holding the array with `{ defaultValue: Ember.A() }` for it to work.

devinus:
@slindberg Disregard. This might have been weird behavior with Ember Data (that's been changed) that I was relying on.

devinus:
@slindberg I'm going to confirm that everything is working good so far in our application. :+1: :100: :tada:

slindberg:
Awesome, thanks a ton @devinus :beers:

I want to get @igorT's take on this PR before merging though, so hang tight for a release.

bower.json
```diff
@@ -18,11 +18,12 @@
     "testem.json"
   ],
   "dependencies": {
-    "ember-data": "1.0.0-beta.17"
+    "ember-data": "1.13.5"
```
devinus: This should allow for higher than `1.13.5` like the recently tagged `1.13.6`.

pifantastic: Yep, a `~1.13.5` would be nice here since minor changes are being released frequently right now.

Kilowhisky:
Whats the status on this?

devinus:
Any chance we can get a tag soon?

packages/model-fragments/lib/fragments/ext.js
```diff
@@ -152,83 +100,141 @@ Model.reopen({
...
+function getFragmentTransform(container, store, attributeType) {
+  var registry = container._registry || container;
+  var containerKey = 'transform:' + attributeType;
+  var match = attributeType.match(/^-mf-(fragment|fragment-array|array)(?:\$([^$]+))?(?:\$(.+))?$/);
+  var transformType = match[1];
+  var modelName = match[2];
+  var polymorphicTypeProp = match[3];
+
+  if (!registry.has(containerKey)) {
+    var transformClass = container.lookupFactory('transform:' + transformType);
```
kmiyashiro: This is returning `undefined` in serializer test for a model that uses fragments. I'm not sure why it's not finding it, as if the initializer is not being run for registering the transform. `transformType` is `fragment`.

I'm using ED 1.13.8 and Ember 1.13.7

kmiyashiro: Oh, could it be that it's failing because the npm package is still 0.3.3 but the bower is set to this one? Not sure how ember-cli testing uses bower vs npm packages.

DingoEatingFuzz:
:tada: :beers:

slindberg:
Sorry for the delay all, released in [v0.4.0](https://github.com/lytics/ember-data.model-fragments/tree/v0.4.0).

workmanw:
Yay! Thank you! I'm working this weekend to upgrade to the latest ember-data. I'll report any issues I find.

jackweinbender:
Thanks for all your great work!

marius0:
Thank you for 0.4.0!

Quick remark though, I still see "0.3.3" in the console when starting the app.

DEBUG: Model Fragments   : 0.3.3+8b1fcdd5
ember.debug.js:5378DEBUG: -------------------------------

slindberg:
@marius0 thanks, I'll fix in next minor release

packages/model-fragments/lib/fragments/ext.js
slindberg: @kmiyashiro Yeah, that could definitely cause that error. There are [shims](https://github.com/lytics/ember-data.model-fragments/tree/v0.4.0/ember-addon/app/transforms) for including the appropriate transforms into your application, and they are included by the NPM package. Can you try again using the v0.4.0 tag?

kmiyashiro: @slindberg I'm getting the same error with 0.4.0. Do I have to manually register the transforms in an initializer?

slindberg: No, you should not need to do anything special. As long as both model fragments packages are at `0.4.0` in `package.json` and `bower.json`, ember-cli should include [shims](https://github.com/lytics/ember-data.model-fragments/tree/v0.4.0/ember-addon/app/transforms) in your app tree, which then get picked up by the resolver. I will investigate further.

kmiyashiro: Just to be clear, my app works fine in development, it's just the serializer unit test that's failing. It's just the default generated serializer test that ember-cli generates, and I added the models to `needs`. In this case, all models except `account` are fragments.

```js
import { moduleForModel, test } from 'ember-qunit';

moduleForModel('order', 'Unit | Serializer | order', {
  // Specify the other units that are required for this test.
  needs: [
    'model:account',
    'model:address',
    'model:discount',
    'model:line-item',
    'model:payment-source',
    'serializer:order'
  ],
});

test('it serializes records', function(assert) {
  var record = this.subject();

  var serializedRecord = record.serialize();

  assert.ok(serializedRecord);
});
```

bower.json
```diff
@@ -18,11 +18,12 @@
     "testem.json"
   ],
   "dependencies": {
-    "ember-data": "1.0.0-beta.17"
+    "ember-data": "^1.13.5"
```
devinus: This should change to allow 2.0.

packages/model-fragments/lib/fragments/ext.js
slindberg: Ah, I was mistaken about not needing to do anything special. The resolver will know how to look up `transform:fragment`, but you still need to explicitly add it to the `needs` array in unit tests. If your model uses a `hasManyFragments` property, you will also need to explicitly include `transform:fragment-array` (or `transform:array`). Does this solve the issue for you?

kmiyashiro: That did it! I guess I never had to include a transform before. Thanks for working through it.

Issue #10, opened 2014-05-21
Allow fragment attribute setters to accept object/array literals


slindberg:
It may be convenient to avoid the step of calling `store#createFragment` before assigning a new fragment to a `DS.hasOneFragment` or `DS.hasManyFragments` property, e.g.

```js
// 'names' is a `DS.hasManyFragments` property
record.set('names', [ { first: 'Rob', last: 'Stark' } ]);

// 'names' is a `DS.hasOneFragment` property
record.set('name', { first: 'Rob', last: 'Stark' });
```

vs.

```js
// 'names' is a `DS.hasManyFragments` property
record.set('names', [
  store.createFragment('name', { first: 'Rob', last: 'Stark' })
]);

// 'names' is a `DS.hasOneFragment` property
record.set('name', store.createFragment('name', { first: 'Rob', last: 'Stark' }));
```

Implications TBD.

PR #111, opened 2015-07-16
deep copy fragments

christophersansone:
I believe the expected behavior for copying fragments would be to clone the entire fragment tree.  Right now `fragment.copy()` does not handle nested fragments this way.  My approach is simple: serialize the fragment to make the deep copy, then create a new fragment and deserialize it.

### DIFF
--- a/packages/model-fragments/lib/fragments/model.js
+++ b/packages/model-fragments/lib/fragments/model.js
@@ -170,13 +170,10 @@ var ModelFragment = CoreModel.extend(Ember.Comparable, Ember.Copyable, {
   copy: function() {
     var store = get(this, 'store');
     var type = store.modelFor(this.constructor);
-    var data = {};
-
-    // TODO: handle copying sub-fragments
-    Ember.merge(data, this._data);
-    Ember.merge(data, this._attributes);
-
-    return this.store.createFragment(type, data);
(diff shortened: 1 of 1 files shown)


slindberg:
Thanks for the PR! You are correct that the expected behavior is for the entire fragment tree to be copied, and it's currently on my [TODO list](https://github.com/lytics/ember-data.model-fragments/issues/110) for the next minor (or major) release.

Unfortunately I don't think serializing the fragment is the right thing to do here, because `setupData` requires normalized data, and `.serialize` passes data through custom serializers. This means that the data would have to get re-normalized before being passed to `setupData`, which adds an unnecessary step. I think a better approach is to iterate over fragments and copy them recursively.

Also note that changing this method on a branch based off master will conflict with the [v1.13 compatibility branch](https://github.com/lytics/ember-data.model-fragments/pull/109).

christophersansone:
Thanks for the response @slindberg!  Agreed that this is the ultra-simplistic and non-performant approach but thought it may be a working solution that could be optimized later when #10 is finished.

Thanks for all your hard work.  This addon is definitely a piece of the puzzle missing from Ember Data!

louy:
#114 is an alternative.

slindberg:
Closing in favor of #114.

PR #120, opened 2015-08-17
fix copying fragment properties

jakesjews:


slindberg:
Guess I had poor test coverage there. Thanks!

PR #123, opened 2015-08-18
Copy nested fragments

### LINKED ISSUE #98: Unable to copy a fragment
I'm trying to clone a DS.Model instance. When I use get and set for copying fragments I get the following error:
> "Assertion Failed: Fragments can only belong to one owner, try copying instead"

If I use `Ember.copy` I get this:
> "Assertion Failed: Cannot clone an Ember.Object that does not implement Ember.Copyable"

What is the correct way to achieve what I want?

louy:
Closes #98.


louy:
Closing in favor of #125.

Issue #110, opened 2015-07-12
Release v1.0 checklist


slindberg:
Possibly v1.13?

- [x] Ember Data v1.13.x compatibility #99 
- [x] Stop using Ember's prototype extensions #104 
- [x] Allow for custom deserializing in transform #95 
- [x] Change default value of `DS.hasManyFragments` to be empty array #36, #40
- [x] Add ability to pass normalized JSON to `createFragment`, fragment attributes #10, #105
- [x] Support nested fragments when copying #98
- [x] Reset fragment owners when removed from a record #133
- [x] Use `DS.JSONSerializer` as default fragment serializer #135
- [x] Fix deprecation warnings (`Ember.create`, etc.)
- [x] Export `MF` global and create cli shim
- [x] Deprecate `DS.hasOneFragment` in favor of `MF.fragment`, `DS.hasManyFragments` for `MF.fragmentArray` & `MF.array` #11, #13 
- [ ] ~~Investigate record relationships #18~~

knownasilya:
What's the status on this?

karellm:
@slindberg Could you tell us what is going on? I think this is the only library in our project preventing us to move forward to `1.13` and `2.0`. Thanks!

slindberg:
I decided to make this list the target for a 1.0 release, and added ED 1.13 compatibility in v0.4.0

jackweinbender:
Is there any plan to make this work with the JSONAPI Serializer? It occurred to me that the JSONAPI spec supports  "any valid JSON value" for its attributes property. Assuming the ED team isn't planning on making this part of the Serializer, it seems like a good gap for EDMF to fill.

http://jsonapi.org/format/#document-resource-object-attributes

jackweinbender:
Is there any plan to allow use with JSONAPI Serializer? I was noticing that [the JSONAPI spec states][1] that "Complex data structures involving JSON objects and arrays are allowed as attribute values," Which—since ED doesn't seem to support this—might be a nice lacuna for EDMF to fill.

[1]: http://jsonapi.org/format/#document-resource-object-attributes

slindberg:
[...]

PR #127, opened 2015-08-20
hasManyFragments now defaults to an empty array

### LINKED ISSUE #40: Default value for `hasManyFragments`
Currently when a `defaultValue` option isn't specified for a `hasManyFragments` attribute, [it is set to `null`](https://github.com/lytics/ember-data.model-fragments/blob/dade27ee147902794dc837058b6f608496fd30e4/packages/model-fragments/lib/fragments/attributes.js#L241). While `null` is a valid value of the property, it is more intuitive that the default is an empty fragment array, and is more consistent with how `DS.hasMany` works already. ~~In fact there currently is no way to force the creation of a fragment array once a record has been created.~~ However this is a minor breaking change, so should probably slated for a future major version bump.

Refs #36.

EDIT: a fragment array property can be set explicitly to an array-like object, which can either be empty or contain valid fragment models.

christophersansone:
Fixes #40.


christophersansone:
Love to get some feedback on the tests.  I couldn't quite tell which ones should be specifically testing for `null` and which ones should be testing for the default empty array.  I tend to think we should assume the default empty array in most cases and then test for the null default value as an edge case.

slindberg:
Thanks for doing the work on this one :beers:

Since `null` is still a valid value for fragment arrays, those tests need to stick around. The rollback-to-empty-array case _should_ be covered by other tests, but it couldn't hurt to add an extra one.

packages/model-fragments/lib/fragments/attributes.js
```diff
@@ -319,13 +319,15 @@ function getDefaultValue(record, options, type) {
 
   if (typeof options.defaultValue === "function") {
     value = options.defaultValue();
-  } else if (options.defaultValue) {
+  } else if ("defaultValue" in options) {
```
slindberg: I'm not a huge fan of `in`, can you change this to `options.defaultValue !== undefined`.

packages/model-fragments/tests/integration/fragments/dependent_state_test.js
```diff
@@ -596,7 +596,7 @@ test("a `DS.hasManyFragments` fragment property that is set to null can be rolle
   });
 });
 
-test("a `DS.hasManyFragments` fragment property that is null can be rolled back", function() {
+test("a `DS.hasManyFragments` fragment property that is empty can be rolled back", function() {
```
slindberg: These tests shouldn't change, since `null` is still valid.

slindberg:
Also, since this will get merged as part of the 1.0.0 release, it should remove the warning about it changing and related tests in `deprecation_test.js`.

christophersansone:
@slindberg Thanks for the feedback, I'll make those changes.  I originally made the changes before #126 was merged and became unsure how to handle it all afterwards.  I think what I'll do is remove the warning, remove `defaultValue: []` in the tests, arrange the tests to assume the empty array default, and then make sure there is test coverage specifically for `defaultValue: null` cases.  Agreed?

Also, I'm thinking it would help have a v1.0 branch.  As each PR for #110 is complete, it can be merged into v1.0 immediately.  If contributors just pull from master each time, we'll likely run into merge conflicts.

christophersansone:
I made the changes as discussed above.

christophersansone:
@slindberg I noticed a small inconsistency I thought I'd point out.  According to one of the tests, when a `hasManyFragments` property is initialized to `null`, then the value should remain `null` instead of using the default value.  This test was actually passing for the wrong reason: it was selecting the default value for any falsy value, but the default value was also null.

I made a change to `hasManyFragments` to use the default value only if the value is specifically `undefined`.  This will now allow nulls to remain null regardless of the default value.  However, this is not the case with `hasOneFragment`: any falsy value will use the default value.  Is that the expected behavior for both `hasOneFragment` and `hasManyFragments`?

Incidentally, I checked how Ember Data applies default values.  If the key exists in the data (using the `in` operator), it uses the value.  If the key does not exist, it uses the default value.  Personally, I'd prefer being as consistent with ED as possible: use the `in` operator and apply it to both `hasOneFragment` and `hasManyFragments`.  I'd be happy to make the change if you'd like.

slindberg:
Sorry for the long delay! Summertime :smile: 

> This test was actually passing for the wrong reason: it was selecting the default value for any falsy value, but the default value was also null.

Good catch. To make sure we're talking about the same thing, you mean when a fragment property has a falsy value in the payload, the default value always clobbers it, correct?

> Is that the expected behavior for both hasOneFragment and hasManyFragments?

Nope, this is definitely a bug.

> Personally, I'd prefer being as consistent with ED as possible

Agreed. This should be fixed independent of this PR (then rebase this one on top of it), sounds like you're up for tackling that too?

christophersansone:
@slindberg Ha, no worries for the delay, except that my availability has dried up!  I'm hoping to get to this by the end of the week.

christophersansone:
Created #138 to address the default value issue.  PR coming shortly.

PR #139, opened 2015-10-02
Established values are no longer clobbered by defaultValue

### LINKED ISSUE #138: Falsy values are getting clobbered by defaultValue
In Ember Data (1.13.x and 2.0), the defaultValue defined on a `DS.attr` is only applied if the value is completely unspecified: the data from the server did not contain the key, and the property has not been set in the client.  Its determination is: `if (key in data) ... else ...`.

Currently, this library works a bit differently.  If a value is falsy at all (`false`, `null`, `undefined`, etc.), it will use the default value.  Its determination is: `data[key] || defaultValue`.  As a result, actual intended values may get unintentionally clobbered by the default value.

Because the intention of this library is to look and feel as similar to Ember Data as possible, the expectation should be for default values to behave consistently in both libraries.

christophersansone:
Fixes #138.  Added some basic `defaultValue` tests as well.

Once this is merged into master, I'll rebase #127, and hopefully we can get that one in too.

### DIFF
--- a/packages/model-fragments/tests/unit/default_value_test.js
+++ b/packages/model-fragments/tests/unit/default_value_test.js
@@ -1,0 +1,87 @@
+var env, store, Person;
+
+QUnit.module("unit/fragments - Default Values", {
+  setup: function() {
+    Person = DS.ModelFragment.extend({
+      firstName : DS.attr("string", { defaultValue: 'Bob' }),
+      lastName  : DS.attr("string"),
+      title     : DS.attr("string", { defaultValue: null }),
+      company   : DS.attr("string", { defaultValue: undefined }),
+      employed  : DS.attr("boolean", { defaultValue: false })
+    });
+
+    env = setupEnv({
+      person: Person,
+    });
+
+    store = env.store;
+  },
+
+  teardown: function() {
+    env = null;
+    store = null;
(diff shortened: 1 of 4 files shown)


christophersansone:
Just realized I needed to write some tests specifically for `DS.hasOneFragment` and `DS.hasManyFragments`.  Stay tuned, probably tomorrow.

packages/model-fragments/tests/unit/default_value_test.js
```diff
@@ -0,0 +1,87 @@
...
+    attributes: {}
+  });
+
+  return store.find('person', 1).then(function(person) {
+    person.set('firstName', null);
+    equal(person.get('firstName'), null, "default value is not used when the value is set to null");
+    person.set('firstName', undefined);
+    equal(person.get('firstName'), undefined, "default value is not used when the value is set to undefined");
+  });
+});
```
slindberg: These tests appear to only be testing the behavior of `DS.attr`, which should have coverage as part of ED's test suite. Also, `Person` extends from `DS.ModelFragment`, yet these tests are pushing person records into the store with an ID?

slindberg:
Awesome, thanks for doing this!

PR #143, opened 2015-10-05
Use public registry.

### LINKED ISSUE #141: Ember 2.1 Compatability
[This](https://github.com/lytics/ember-data.model-fragments/blob/master/packages/model-fragments/lib/fragments/ext.js#L244) is raising an exception that `registry.has is not a function`.

Replacing the definition of `registry` with
```
var registry = container.registry || container._registry || container;
```
fixes the exception.

Also, less critical, but uses of `Ember.warn` are raising deprecation notices.

abuiles:
Since Ember 2.1 `registry` is public.

See emberjs/ember.js#11440.

Closes #141


slindberg:
:beers:

PR #148, opened 2015-10-12
Changed default fragment serializer to use store default

### LINKED ISSUE #135: Use DS.JSONSerializer as default fragment serializer
As [@dfreeman pointed out](https://github.com/lytics/ember-data.model-fragments/issues/110#issuecomment-142386315), if the application serializer extends from `DS.JSONAPISerializer`, a custom serializer must be created for every fragment that extends from `DS.JSONSerializer` instead. Seems like the best/only way to handle this is to override `Store#serializerFor`.

slindberg:
Closes #135

This changes the behavior of `store.serializerFor` to skip `serializer:application` and the default adapter's `defaultSerializer` property. It will also use `serializer:-fragment` if registered as a convenience for apps that want a serializer that is different from the application serializer, but do not want to create a boilerplate serializer class for every fragment class. Prefixing it with a dash requires devs to register it in a custom initializer, but it avoids potential naming clashes.

This is a breaking change since many apps may rely on using the application serializer as the default fragment serializer, and will be part of the 1.0 release.

cc @dfreeman

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,8 @@ App.NameSerializer = DS.JSONSerializer.extend({
 });
 ```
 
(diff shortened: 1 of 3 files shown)


dfreeman:
:+1: Looks great – thanks for knocking this out!

PR #149, opened 2015-10-13
update initializer for ember 2.1

jakesjews:


packages/model-fragments/lib/initializers.js
```diff
@@ -7,7 +7,8 @@ var initializers = [
     name: "fragmentTransform",
     before: "store",
 
-    initialize: function(container, application) {
+    initialize: function() {
+      var application = arguments[1] || arguments[0];
```
slindberg: Can you use the named arguments instead? i.e.

```js
if (!application) {
  application = container;
}
```

And add a comment noting that it's for Ember 2.1 compat? Thanks!

jakesjews: no problem

PR #150, opened 2015-10-13
Added `MF` namespace, updated API, added deprecations

### LINKED ISSUE #11: Simplify fragment API naming
`DS.ModelFragment`, `DS.hasOneFragment`, and `DS.hasManyFragments` are all mouthfulls. The tradeoff is always clarity vs. brevity, and the current naming may be a little too heavy on the clarity side. Potential simplifications:

`DS.ModelFragment` → `DS.Fragment`
`DS.hasOneFragment` → `DS.fragment`, `DS.fragmentObject`
`DS.hasManyFragments` → `DS.fragments`, `DS.fragmentArray`

 Moving away from likeness to `DS.belongsTo` and `DS.hasMany` may be desirable once all relationship properties return promises, which fragment properties will never do.

### LINKED ISSUE #13: Move 'primitive' support from `DS.hasManyFragments` into a new property type
Arrays of non-fragment types (usually primitives) are currently supported by omitting the fragment type from the `DS.hasManyFragments` definition. This is 1) misleading since primitives are not `DS.ModelFragment` instances, and 2) because a primitive type (and therefore transform) cannot be specified and enforced.

A possible solution is to introduce a `DS.hasManyPrimitives` property type, however this is also misleading since the types this property would support aren't limited to primitives (think `date` type). The distinction here is that there is no need for special state handling on the nested value. This loosely translates to when the value is _serialized_ to a primitive for JSON transport. So then what to name it?

This also potentially impacts simplified naming in #11.

slindberg:
Exports a new global `MF` and updates the model fragments API (as discussed in #11) in preparation for v1.0. Adds deprecation warnings for all outdated APIs

Changes:

`DS.ModelFragment` → `MF.Fragment`
`DS.hasOneFragment` → `MF.fragment`
`DS.hasManyFragments` → `MF.fragmentArray`

Support for non-fragment `hasManyFragment` properties has been removed from `MF.fragmentArray` and replaced with a new property `MF.array`, which supports transforms:

```js
export default DS.Model.extend({
  things: MF.array('string'),
  stuff: MF.array('my-custom-transform')
});
```

[Full API](https://github.com/lytics/ember-data.model-fragments/blob/mf-namespace/packages/model-fragments/lib/main.js#L71-L88)

Closes #11
Closes #13


workmanw:
:+1:

Issue #133, opened 2015-09-10
Reset fragment owners when removed from a record


slindberg:
Fragments cannot belong to more than one record, but currently removing a fragment from a record does not free it from its owner, and it cannot be added to another record, which forces copying of a usable object.

The `hasOneFragment` property should manage resetting the owner of any fragments that are replaced, and fragment arrays should do the same with `replaceContent`.

```js
var Thing = DS.Model.extend({
  config: hasOneFragment('config'),
  widgets: hasManyFragments('widget')
});

var thing = store.peekRecord('thing', 1);
var newThing = store.createRecord('thing');

var config = thing.get('config');
thing.set('config', null); // clears `config`'s owner
newThing.set('config', config); // allowed since `config` now has no owner

var lastWidget = thing.get('widgets.lastObject');
thing.get('widgets').removeObject(lastWidget); // clears `lastWidget`'s owner
newThing.get('widgets').addObject(lastWidget); // allowed since `lastWidget` now has not owner
```

PR #157, opened 2015-10-25
Revert fragment owner reseting

slindberg:
This reverts the behavior described in #133. It turns out there was a very good reason for only ever allowing a fragment to belong to one owner: it prevents conflicts when rolling back a record with fragments:

```js
var fragment = record1.get('fragment');

record1.set('fragment', null); // Previously, this cleared the fragment's owner

record2.set('fragment', fragment); // Possible, since the fragment has no owner

record1.rollback();

fragment.get('owner'); // ???
```

this is a breaking change for anyone that has upgraded to v0.4.2 or v0.4.3 and now relies on this behavior. However, since chances of that are small, and this was an oversight on my part and introduces a bug, this should be treated as like a bugfix (no minor version number bump).

@martinthogersen I'm curious if you've updated to use this behavior yet.

### DIFF
--- a/packages/model-fragments/lib/fragments/array/fragment.js
+++ b/packages/model-fragments/lib/fragments/array/fragment.js
@@ -185,13 +185,6 @@ var FragmentArray = StatefulArray.extend({
     var replacedContent = content.slice(index, index + amount);
     var fragments = normalizeFragmentArray(this, replacedContent, objs);
 
-    // If fragments get removed from the array, clear their owners
-    if (fragments.length < replacedContent.length) {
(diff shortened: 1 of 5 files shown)


martinthogersen:
Yes. The code currently looks like this:

```js
currentSection.get('applicationQuestions').removeFragment(question)
moveSectionTarget.get('applicationQuestions').addFragment(question.copy())
```

I'm using version 0.4.3.

slindberg:
Ok, well since you're using `.copy()`, this won't be a breaking change for you, no need to make any updates.

Issue #160, opened 2015-11-08
Ember Data default serializers cannot be looked up


thec0keman:
Calling `store.serializerFor('application')` will fail with `No model was found for 'application'`.

I believe this is because in the [override](https://github.com/lytics/ember-data-model-fragments/blob/master/packages/model-fragments/lib/fragments/ext.js#L88) version we first call `modelFor`, which is a bit differ of a [path](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1972) than in Ember Data.

slindberg:
Thanks for reporting this :beers:, fixed in #162 release to follow soon.

PR #161, opened 2015-11-08
Fix for Ember Data default serializers

thec0keman:
Fix for [#160](https://github.com/lytics/ember-data-model-fragments/issues/160)

(I suspect there is a better path to solving this issue).


slindberg:
Closing in favor of #162. I pulled your commit into that branch though, so you should get contributor credit :koala:

BTW, you should consider changing the author email in your git config to one that is registered with Github so your profile can be linked to from commits.

thec0keman:
Good point, thanks!

PR #171, opened 2015-12-23
Fixed missing `data` envelop in store.push calls

slindberg:


workmanw:
:+1: . FWIW It seems that our app is working as expected with ember-data 2.2 and ember-data-model-fragments#master. Our tests pass and basic QA seems good.

slindberg:
Yeah, I don't think I'm going to create a v2.2 release since all tests pass w/ Ember/ED v2.2.0. Thanks for the info :smile:

workmanw:
Seems reasonable to me. Maybe update the README.md? I can submit a PR, but it may be just as easy for you to edit it.

slindberg:
The last line in the compatibility table indicates that ED v2.2 (>= v2.1.x) should use MF v2.1.x, so I don't think a special note is really necessary. TBH I doubt anyone looks at that thing unless something fails anyway.

workmanw:
Oh yea. My brain read that wrong. Been staring at semversion too much. :stuck_out_tongue_winking_eye:

PR #175, opened 2016-01-14
Fix memory leak caused by fragments not being destroyed

### LINKED ISSUE #106: A way to destroy fragment
use case:
```js
var person = store.getById('person', '1');

person.set('birthday', myNewBirthday);

var name = store.createFragment('name', {
  first : 'Hugor',
  last  : 'Hill'
});

person.set('name', name);

//then later on

var name = person.get('name');

person.set('name', null);

//what to do with name?
```
I don't want to rollback `person` and lose their birthday change. Should there be a `name.destroyFragment` or something similar to remove it from the store?

dwickern:
I tracked down a memory leak since my acceptance tests started running the process out of memory. This is from a heap snapshot showing the ember `Container` instance being rooted in `window`:

```
container in Class @8620175
 _object in ChainNode @8620223
  _parent in ChainNode @8614719
   [2] in Array @8610299
    isDirty in EmptyObject @8607285
     chains in ChainWatchers @8601471
      _chainWatchers in Meta @8065411
       __ember_meta__ in @7311237
        saved in @8042657
         loaded in @8042653
          model$fragments$lib$fragments$states$$default in system / Context @8042649
           context in model$fragments$lib$fragments$attributes$$fragmentArray() @8042947
            fragmentArray in Class @7447369
             MF in Window
```
I had about as many containers rooted as the number of acceptance tests.

The leak happens because `window.MF` has the `FragmentRootState`. Calling e.g. `fragment.get('isDirty')` registers a watcher on `currentState` which is the rooted state object. `fragment.destroy` never gets called to clean up the watcher.

When you call `application.destroy` or `store.unloadAll`, ember-data goes through its registry and destroys each model instance. There's no such registry for fragments so they hang around forever.

The pull request changes `DS.Model#destroy` to recursively destroy any fragments which the model owns. Ownership is tracked by a new property `_owns` (which is the inverse of `_owner`).

With this fix applied, chrome's `performance.memory.usedJSHeapSize` after running my test suite improves from 624 MB (at edmf master branch) to 239 MB.

Also fixes #106

### DIFF
--- a/packages/model-fragments/lib/fragments/array/fragment.js
+++ b/packages/model-fragments/lib/fragments/array/fragment.js
@@ -234,6 +234,20 @@ var FragmentArray = StatefulArray.extend({
     var fragment = store.createFragment(type, props);
 
     return this.pushObject(fragment);
+  },
+
+  willDestroy: function() {
+    this._super.apply(this, arguments);
+
+    // destroy the current state
+    this.forEach(function(fragment) {
+      fragment.destroy();
+    });
+
+    // destroy the original state
+    this._originalState.forEach(function(fragment) {
+      fragment.destroy();
+    });
   }
 });
 
--- a/packages/model-fragments/lib/fragments/ext.js
+++ b/packages/model-fragments/lib/fragments/ext.js
@@ -7,6 +7,7 @@ import {
   internalModelFor,
   default as Fragment
 } from './fragment';
+import FragmentArray from './array/fragment';
 
 /**
   @module ember-data-model-fragments
@@ -151,6 +152,29 @@ Model.reopen({
 
     return diffData;
   },
+
+  willDestroy: function() {
+    this._super.apply(this, arguments);
+
+    var internalModel = internalModelFor(this);
+    var key, fragment;
+
+    // destroy the current state
+    for (key in internalModel._fragments) {
+      if (fragment = internalModel._fragments[key]) {
+        fragment.destroy();
+      }
+    }
+
+    // destroy the original state
+    for (key in internalModel._data) {
+      if (fragment = internalModel._data[key]) {
+        if (fragment instanceof Fragment || fragment instanceof FragmentArray) {
+          fragment.destroy();
+        }
+      }
+    }
+  }
 });
 
 // Replace a method on an object with a new one that calls the original and then
--- a/packages/model-fragments/tests/integration/nested_test.js
+++ b/packages/model-fragments/tests/integration/nested_test.js
@@ -251,3 +251,51 @@ test("Nested fragments can be copied", function() {
     ok(product !== user.get('orders.firstObject.product'), 'nested fragment copies are new fragments');
   });
 });
+
+test("Nested fragments are destroyed when the owner record is destroyed", function() {
+  store.push({
+    data: {
+      type: 'user',
+      id: 1,
+      attributes: {
+        info: {
+          name: 'Tyrion Lannister',
+          notes: [ 'smart', 'short' ]
+        },
+        orders: [
+          {
+            amount: '10999.99',
+            products: [
+              {
+                name  : 'Lives of Four Kings',
+                sku   : 'old-book-32',
+                price : '10999.99'
+              }
+            ]
+          }
+        ]
+      }
+    }
+  });
+
+  return store.find('user', 1).then(function(user) {
+    var info = user.get('info');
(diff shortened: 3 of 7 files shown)


slindberg:
@dwickern you are my hero :hatched_chick:

Out of curiosity, does removing the `FragmentRootState` export fix the leak? If it does, I'd be super tempted to break just break SemVer and call it a day.

I like your approach, but I wonder about the impact on applications that create/release large numbers of fragment instances during the lifetime of the app. This could be solved by intelligently destroying fragments when they are overwritten by an [adapter](https://github.com/lytics/ember-data-model-fragments/blob/c804081/packages/model-fragments/lib/fragments/fragment.js#L115-L119) [commit](https://github.com/lytics/ember-data-model-fragments/blob/35d81a1e80ab0128a3b04e686ddaa1dc956fad05/packages/model-fragments/lib/fragments/array/fragment.js#L122-L130) (in addition to when the model is destroyed). This might end up being cleaner too, since the model destroy hook could just use the `_fragments` property instead of tracking them separately in `_owns`.

I'll look into why the build failed too, not sure what's going on there.

dwickern:
> Out of curiosity, does removing the FragmentRootState export fix the leak? If it does, I'd be super tempted to break just break SemVer and call it a day.

Probably not. The root state is always going to be referenced somehow. You could create a new state object for each fragment but that invites a new set of problems, and it would still be hard to reason about whether there are leaks.

> I like your approach, but I wonder about the impact on applications that create/release large numbers of fragment instances during the lifetime of the app.

Originally I was going to destroy the fragment when the value gets replaced, but that breaks rollback:
```
var name = person.get('name');
person.set('name', null); // the `name` fragment gets destroyed
person.rollbackAttributes();
person.get('name.isDestroyed') // whoops!
```

You could come up with similar examples where destroying the fragment on commit would not work:
```
var name = person.get('name');
person.set('name', null);
person.commit();

// ... some time later
person.set('name', name);
```

There is no way to know when it's safe to destroy, except when the fragment's owner gets destroyed (given that ownership is permanent).

Probably the best solution to your concern is to make `_owns` a `WeakSet` instead of an array. Thoughts?

slindberg:
> Probably not. The root state is always going to be referenced somehow

Sigh, indeed.

Rollbacks wouldn't be an issue if fragments are only destroyed when changes are committed, but you're second example is perfectly valid. What if instead the store itself kept references to fragments, but only when they become inaccessible from models (commit, push). That way dangling fragments could be destroyed separately from ones referenced by models. It's not terribly different than your approach, but it does avoid the double tracking in models.

I'm also intrigued by the `WeakSet` idea. There are a lot of polyfills out there, ranging from [dirt simple](https://github.com/dfcreative/weakset) to [ridiculously convoluted](https://github.com/zloirock/core-js) to [not actually weak](https://github.com/WebReflection/es6-collections#the-weakmap-is-not-weak--and-why). Do you have a go-to? I would like to backport this fix to the 1.13 branch, which means it has to be IE8 compatible (no `defineProperty`).

Still don't know wtf is going on with the Travis build :angry:

dwickern:
`WeakSet` doesn't help us actually since it's not iterable. I had assumed it worked like other language APIs.

I changed my implementation so there is no additional ownership tracking. I destroy the original and current fragments, but not any other fragments which might have been assigned to the model. I pulled out this test which now fails: dwickern/ember-data.model-fragments@598ce6b6393ff66ece4f6535b7bb83ae9181353f

Here's chrome's reported heap size after running my test suite against each branch:
Master branch: 624 MB
Original PR: 239 MB (dwickern/ember-data.model-fragments@ca8042a80a93e164a6204d013be186b9762e4f11)
Updated PR: 254 MB (dwickern/ember-data.model-fragments@a50bec4b06aa6a09ac76fdb720a17a59ea3c9413)

So we are leaking a bit more, but it's still a massive improvement and this change should eliminate your concerns.

slindberg:
Apologies again for the enormous delay, would you mind rebasing with master now that [travis is working again](https://github.com/lytics/ember-data-model-fragments/pull/181)? I hate to lose all of the work you did previously, but I do think this approach has much lower conceptual overhead, while still addressing the bulk of the problem. I'll see if I can come up with something to destroy those orphaned fragments, but in the meantime lets get this merged :shipit:

dwickern:
There you go. My tests used ES6 `let` so that might have been your "syntax error"

Issue #25, opened 2014-09-24
Handling errors


trobrock:
Is there any way to assign errors to a fragment? We have a api that will return errors for the nested model fragments and would like to be able to show those errors, but I haven't been able to find any good examples of this anywhere.

PR #179, opened 2016-02-12
Fixing #177. Ensures that adapter errors are properly handled.

### LINKED ISSUE #177: Can't rollback fragment after #save failure
Seems to be happening on both object and array fragments. If the adapter's `save` fails, the rollback does cascade down but it doesn't rollback each fragment's attributes.

It's clear that the source of the problem is lack of coverage for `adapterDidError` on the Model but even with that added there would need to be an `inFlight` state which will add a bit of complexity. I can probably work on fixing it but I would like to know the preferred direction to take? Since this use to work fine on `0.4.x` !

Here is a spec for the bug (add to tests/integrations/save_test.js)
```js
test("existing fragments can be rolled back on failed save", function() {
  expect(3);

  var data = {
    name: {
      first: "Eddard",
      last: "Stark"
    },
    addresses: [
      {
        street: "1 Great Keep",
        city: "Winterfell",
        region: "North",
        country: "Westeros"
      }
    ]
  };

  store.push({
    data: {
      type: 'person',
      id: 1,
      attributes: data
    }
  });

  env.adapter.updateRecord = function() {
    return Ember.RSVP.reject({});
  };

  var mrStark, name, address;

  return store.find('person', 1).then(function(person) {
    mrStark = person;

    name = mrStark.get('name');
    address = mrStark.get('addresses.firstObject');

    name.set('first', 'BadFirstName');
    name.set('last', 'BadLastName');
    address.set('street', 'BadStreet');

    return mrStark.save();
  }).catch(function() {
    mrStark.rollbackAttributes();

    equal(`${name.get('first')} ${name.get('last')}`, 'Eddard Stark', "fragment name rolled back");
    equal(address.get('street'), '1 Great Keep', "fragment array fragment correctly rolled back");
  });
});
```

workmanw:
This PR is intended to fix #177. It ensures adapter errors are properly handled.

There are some open questions here and I would like @slindberg to weigh in when he is able to. 

One question is should the fragments call `internalModelFor(this)._saveWasRejected()` instead of `    internalModelFor(this).adapterDidError(error);`? The challenge I discovered was that the internal-model's [adapterDidError](https://github.com/emberjs/data/blob/266ae624a6005092f7e3758005920039081b8884/addon/-private/system/model/internal-model.js#L735-L743) ultimately tries to call `this.send('becameError');`. However since the fragment never enters the inFlight state, that `becomeError` isn't a valid action. 

Second question, what does it mean for a fragment to be in "error"? I passed the error object for the model through all the fragment hooks so it would be available, but they really don't consume them. If we wanted fragments to consume and reflect the error it would create more questions. For example, "how should a fragment use an error?", "What the format of that error would be?", etc, etc, etc. 

I spent several hours thinking about this. I feel like this is the most pragmatic solution. While it has it's shortcomings, it does fix the bug and restore what I would say is expected functionality.

### DIFF
--- a/packages/model-fragments/lib/fragments/array/fragment.js
+++ b/packages/model-fragments/lib/fragments/array/fragment.js
@@ -130,6 +130,20 @@ var FragmentArray = StatefulArray.extend({
   },
 
   /**
+    @method _adapterDidError
+    @private
+  */
+  _adapterDidError: function(error) {
+    this._super(error);
+
+    // Notify all records of the error; if the adapter update did not contain new
+    // data, just notify each fragment so it can transition to a clean state
+    this.forEach(function(fragment) {
+      fragment._adapterDidError(error);
+    });
+  },
+
+  /**
     If this property is `true`, either the contents of the array do not match
     its original state, or one or more of the fragments in the array are dirty.
 
--- a/packages/model-fragments/lib/fragments/array/stateful.js
+++ b/packages/model-fragments/lib/fragments/array/stateful.js
@@ -119,6 +119,10 @@ var StatefulArray = Ember.ArrayProxy.extend(Ember.Copyable, {
     }
   },
 
+  _adapterDidError: function(/*error*/) {
+    // No-Op
+  },
+
   /**
     If this property is `true` the contents of the array do not match its
     original state. The array has local changes that have not yet been saved by
--- a/packages/model-fragments/lib/fragments/ext.js
+++ b/packages/model-fragments/lib/fragments/ext.js
@@ -274,6 +274,18 @@ decorateMethod(InternalModelPrototype, 'adapterDidCommit', function adapterDidCo
   }
 });
 
+decorateMethod(InternalModelPrototype, 'adapterDidError', function adapterDidErrorFragments(returnValue, args) {
+  var error = args[0] || create(null);
+  var fragment;
+
+  // Notify fragments that the record was committed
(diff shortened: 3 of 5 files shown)


jakesjews:
I tried pulling these changes into our application and all our tests pass so this gets my approval. The Travis log does show one jshint error `jshint/model-fragments/lib/fragments/fragment.js: line 124, col 30, 'error' is defined but never used` so we probably want to fix that up (we really need to get CI working).

workmanw:
@jakesjews Oops. Good catch! I'll fix that up. Thank you.

slindberg:
LGTM, thanks for taking the lead on this one. Sounds like you fell down the error handling rabbit hole thinking about this. I fell in once a year ago trying to address #25, and have been [avoiding it ever since](https://github.com/lytics/ember-data-model-fragments/issues/112). I'm definitely in favor of fixing the blocker and punting on the rest.

Can you rebase with master now that [travis is fixed](https://github.com/lytics/ember-data-model-fragments/pull/181)?

workmanw:
@slindberg Rebased! Thank you.

slindberg:
Looks like updating to PhantomJS 2.x updated the javascript engine to support `let`, but not template strings. Running locally, tests pass in chrome but not phantom. I'll try to run tests through Babel in the 2.3 update, but for now can you strip your template strings out of the tests so I can merge this?

workmanw:
Done.

PR #182, opened 2016-02-25
Fixed issue with 'isNew' for newly created fragments

### LINKED ISSUE #159: isNew detection for new Fragments
In 2.0, is there a way to detect if a fragment is new? I assume there is since if you rollback a fragment when it's new, it will be marked as deleted.

slindberg:
Fixes #159 

@workmanw I'm thinking I just left this out when initially creating the state machine, but I wanted a gut check from you. There's no deleted state for a fragment, but there is a new state, and I don't think there's any harm in reporting it as 'isNew'.

### DIFF
--- a/packages/model-fragments/lib/fragments/states.js
+++ b/packages/model-fragments/lib/fragments/states.js
@@ -107,6 +107,8 @@ var FragmentRootState = {
     created: {
       isDirty: true,
 
+      isNew: true,
+
       setup: dirtySetup,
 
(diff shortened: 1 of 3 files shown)


workmanw:
@slindberg Yeap! Seems straightforward to me. :+1:

jakesjews:
@workmanw are we good to merge this? It LGTM.

workmanw:
Yeap, I think so! I tested this and it worked for me!

PR #183, opened 2016-03-04
Use getOwner to avoid deprecations in newer versions of ember.

cibernox:
I'm getting **loads** of deprecations from this addon because accessing the container has been deprecated since Ember 2.3, in favour of `Ember.getOwner`.

Using the getOwner polyfill this can be cross version.

### DIFF
--- a/packages/model-fragments/lib/fragments/ext.js
+++ b/packages/model-fragments/lib/fragments/ext.js
@@ -8,6 +8,7 @@ import {
   default as Fragment
 } from './fragment';
 import FragmentArray from './array/fragment';
+import getOwner from '../util/ember-getowner-polyfill-polyfill';
 
 /**
   @module ember-data-model-fragments
@@ -46,7 +47,7 @@ Store.reopen({
 
     Ember.assert("The '" + type + "' model must be a subclass of MF.Fragment", Fragment.detect(type));
 
-    var internalModel = new InternalModel(type, null, this, this.container);
+    var internalModel = new InternalModel(type, null, this, getOwner(this));
 
     // Re-wire the internal model to use the fragment state machine
     internalModel.currentState = FragmentRootState.empty;
@@ -300,7 +301,7 @@ JSONSerializer.reopen({
   */
   transformFor: function(attributeType) {
     if (attributeType.indexOf('-mf-') === 0) {
-      return getFragmentTransform(this.container, this.store, attributeType);
+      return getFragmentTransform(getOwner(this), this.store, attributeType);
     }
 
     return this._super.apply(this, arguments);
@@ -308,25 +309,24 @@ JSONSerializer.reopen({
 });
 
 // Retrieve or create a transform for the specific fragment type
-function getFragmentTransform(container, store, attributeType) {
-  var registry = container._registry || container.registry || container;
+function getFragmentTransform(appInstance, store, attributeType) {
   var containerKey = 'transform:' + attributeType;
   var match = attributeType.match(/^-mf-(fragment|fragment-array|array)(?:\$([^$]+))?(?:\$(.+))?$/);
   var transformName = match[1];
   var transformType = match[2];
   var polymorphicTypeProp = match[3];
 
-  if (!registry.has(containerKey)) {
-    var transformClass = container.lookupFactory('transform:' + transformName);
+  if (!appInstance.hasRegistration(containerKey)) {
+    var transformClass = appInstance._lookupFactory('transform:' + transformName);
 
-    registry.register(containerKey, transformClass.extend({
+    appInstance.register(containerKey, transformClass.extend({
       store: store,
       type: transformType,
       polymorphicTypeProp: polymorphicTypeProp
     }));
   }
 
-  return container.lookup(containerKey);
+  return appInstance.lookup(containerKey);
 }
 
 export { Store, Model, JSONSerializer };
--- a/packages/model-fragments/tests/integration/app_test.js
+++ b/packages/model-fragments/tests/integration/app_test.js
@@ -3,7 +3,7 @@ var app;
 QUnit.module("integration - Application");
 
 test("the model fragments initializer causes no deprecations", function() {
-  expectNoDeprecation();
+  ok(true); // expectNoDeprecation();
 
   Ember.run(function() {
     app = Ember.Application.create();
--- a/packages/model-fragments/lib/fragments/transforms/array.js
+++ b/packages/model-fragments/lib/fragments/transforms/array.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Transform from 'ember-data/system/transform';
 import map from '../../util/map';
+import getOwner from '../../util/ember-getowner-polyfill-polyfill';
 
 /**
   @module ember-data-model-fragments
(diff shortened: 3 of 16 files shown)


workmanw:
I was going to waiting until we got this "addon-ifyed" to try to fix those deprecation issues (https://github.com/lytics/ember-data-model-fragments/issues/174). But seeing as this is so simple, it get's a :+1: from me.

jakesjews:
I'm definitely ok porting this over to my fork.

cibernox:
I've found a problem. For some reason I don't understand I'm getting `missing module import from model-fragments/lib/fragments/ext.js for path: ember-getowner-polyfill`. I don't see why, since the addon is installed. I tried moving from `devDependencies` to `dependencies` but no change.

Any idea why I can't import the polyfill? The structure of this addon is different to most addons I've seen.

packages/model-fragments/lib/fragments/ext.js
```diff
@@ -308,25 +309,24 @@ JSONSerializer.reopen({
 });
 
 // Retrieve or create a transform for the specific fragment type
-function getFragmentTransform(container, store, attributeType) {
-  var registry = container._registry || container.registry || container;
+function getFragmentTransform(appInstance, store, attributeType) {
```
cibernox: `getOwner` returns an `appInstance` that is like a mix of the container and the registry, so no need of accessing private apis.

packages/model-fragments/lib/fragments/ext.js
```diff
@@ -308,25 +309,24 @@ JSONSerializer.reopen({
...
   var containerKey = 'transform:' + attributeType;
   var match = attributeType.match(/^-mf-(fragment|fragment-array|array)(?:\$([^$]+))?(?:\$(.+))?$/);
   var transformName = match[1];
   var transformType = match[2];
   var polymorphicTypeProp = match[3];
 
-  if (!registry.has(containerKey)) {
-    var transformClass = container.lookupFactory('transform:' + transformName);
+  if (!appInstance.hasRegistration(containerKey)) {
+    var transformClass = appInstance._lookupFactory('transform:' + transformName);
```
cibernox: except `_lookupFactory`, which for some reason is still private.

jakesjews:
Actually come to think of it, this might not support pulling in the polyfill addon until we finish converting the code to be a normal addon. For the time being we might need to pull in the code from `ember-getowner-polyfill` into the util directory and import that https://github.com/lytics/ember-data-model-fragments/tree/master/packages/model-fragments/lib/util.

cibernox:
I'll do than then.

jakesjews:
@cibernox sorry about the trouble. Once the addon conversion is finished then it should be better.

jakesjews:
@workmanw if we both approve this and CI passes does that mean we are good to merge? I remember @slindberg mentioned we can merge any PRs at least two of us approve.

workmanw:
Sorry for the delay, we had a production issue that swallowed my afternoon. :(

I definitely think we should pull the `ember-getowner-polyfill` code into the repo for now and cut a new version. To me, this seems to be an issue of addon maintenance, rather than feature expansion. I think if @cibernox can make that change and get CI to pass, we can merge.

workmanw:
@jakesjews I think we should try to get this PR merge and then cut a new version. Once that is done, I can help with the effort to complete the addon work. It seems like having a new version pre-addon will offer future ember-data compatibility for a while. That way if there is instability converting to the addon style, users won't be immediately hosed. 

@cibernox I certainly don't want to steal your PR, but if you don't presently have the bandwidth I'd be happy to help finish it up. Let me know what you think.

cibernox:
I do have time, I was just pulling the hair out of my head because tests fail locally for me (but not in travis) and I can't figure out why.

Not in this branch, but in master. It makes no sense.

jakesjews:
@workmanw I totally agree about a pre-addon release. @slindberg how should we go about cutting a new release?

@cibernox I can take a look at the tests whenever I get some free time.

workmanw:
When I run the tests locally from master the only failures I get are about deprecation warnings relating to the container / registry. If you globally find / replace all of the `expectNoDeprecation()` with `ok(true)` all the tests pass for me. I'm using `ok(true)` because many of the tests have expected assert counts.

packages/model-fragments/tests/integration/app_test.js
```diff
@@ -3,7 +3,7 @@ var app;
 QUnit.module("integration - Application");
 
 test("the model fragments initializer causes no deprecations", function() {
-  expectNoDeprecation();
+  ok(true); // expectNoDeprecation();
```
cibernox: Shall I undo this?

workmanw: Yeap :). Now that you've brought the `getOwner` polyfill in, I would expect that if you undid that change for the tests, then reapplied your container/registry => getOwner changes, it should work. I can't see the commit anymore, but are two usages of registry ([both usages](https://github.com/lytics/ember-data-model-fragments/blob/557f4a288a6966fb4e8df48a80871ae11a17eea0/packages/model-fragments/lib/fragments/ext.js#L319-L327) and two usages of container ([1](https://github.com/lytics/ember-data-model-fragments/blob/557f4a288a6966fb4e8df48a80871ae11a17eea0/packages/model-fragments/lib/fragments/ext.js#L320-L329) and [2](https://github.com/lytics/ember-data-model-fragments/blob/master/packages/model-fragments/lib/fragments/transforms/array.js#L64)).

cibernox: Actually there is deprecations coming from somewhere else. Probably directly from ember-data itself.

workmanw: It's possible that it's also the test suite. See: https://github.com/lytics/ember-data-model-fragments/blob/master/tests/ember-setup.js . Likely that will have to be converted to this to using owner too.

cibernox: Trying to convert it. Is less straightforward than the previous change. Will try again at home.

workmanw: Alright, happy to take a swing at it if you want. We had to build a fake owner in our application's unittest suite to replace the container and registry. We basically followed this: 

```
import FakeOwner from 'ember-getowner-polyfill/fake-owner';
  
export default function buildOwner() {
   let registry = new Ember.Registry();
   let container = new Ember.Container(registry);
   return new FakeOwner({ container: container });
}
```

slindberg:
@cibernox thanks for putting this in, I realize what a pain it must be for apps that are trying to stay current with Ember. After talking with @workmanw and @jakesjews yesterday, I'd like to just push forward with #187 and create a 2.3.0 release that requires Ember/ED >= 2.3.0. This way, we won't have to deal with the `getOwner` polyfill at all, and if there's a critical bugfix that affects the 2.1.x version I can just backport it for people stuck on it. Does that sound good?

cibernox:
Absolutely. I've actually run out of cycles for this since I leave on a trip tomorrow. 

I can stay in ED 2.2 a a couple more weeks without problem, that is not preventing me from updating Ember itself.

workmanw:
I'm :+1: on closing this then.

@cibernox I did try to cherry-pick your commits from this PR into the new addonify PR. It become too hairy because we had also refactored the file structure. So I ended up manually fixing the `owner` issue on that branch using your commits as a reference. I do plan to credit you for this in the Changelog. Thanks :).

PR #184, opened 2016-03-04
Lock ember-data to supported version

cibernox:
With the current version strategy in bower.json, this addon will install 2.4.1 and therefore tests will fail in master.

Now it will stay in 2.2.X.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,7 +19,7 @@
   ],
   "dependencies": {
     "ember": "^2.2.0",
-    "ember-data": "^2.2.0"
+    "ember-data": "~2.2.0"
   },
   "devDependencies": {
     "jquery": "^2.0.0",


workmanw:
:+1: I was thinking about this in the context of your `getOwner` PR. This definitely makes sense to me.

cibernox:
Green

cibernox:
This is weird. Tests are green in travis but in my local machine there is a lot of failures. How is the travis environment different from mine?.

I updated my local phantomjs to 2.1 in case that was the difference, but it's not.

Ideas?

cibernox:
Any feedback on this?

jakesjews:
@workmanw I'm good with this are you ok with me merging?

workmanw:
Yeap, this seems okay to me. We should obviously bump this version as we work to support new versions of ember-data (IE 2.3 and 2.4).

PR #186, opened 2016-03-21
WIP - Convert to Ember-CLI Addon

jakesjews:
There is still a bit of cleanup to do but I wanted to get this out in the open. So far this seems to be working when I pull it into my company's applications. The tests are all passing locally but most of them are making deprecated calls to application.\_\_container\_\_. I've only really used mocha for testing ember apps so if anyone who's better at QUnit has some suggestions that would be very helpful.

TODO:
- [ ] Replace deprecated calls to application.\_\_container\_\_ in the tests
- [ ] Replace moduleForAcceptance usage in unit tests (if possible)

Issue #174, opened 2016-01-12
Ember Data 2.3 compatibility


devinus:
`DS` is no longer a global.

devinus:
http://emberjs.com/blog/2016/01/12/ember-data-2-3-released.html

slindberg:
Yep, this gonna be interesting.

dwickern:
Does this mean you can remove the re-exported ember-data classes like `DS.Store`? Chrome opens your store.js instead of ember-data's which makes debugging it quite difficult.

slindberg:
@dwickern that's bugged me for a long time too :stuck_out_tongue: The update will likely fix the issue.

dfreeman:
There's also the fun fallout from Ember 2.3's `this.container` deprecation...

htmldrum:
Any guidance for nubs who just end up crashing and burning on Uncaught ReferenceError: DS is not defined?

kmiyashiro:
rollback to ember-data 2.2

jakesjews:
We were able to upgrade by keeping ember-data in our bower.json file and setting it to 2.3.3 in both bower.json and package.json. You'll see a warning from ember-cli but you can ignore it.

kmiyashiro:
@jakesjews did you check to see if ember-data is included twice in the build?

jakesjews:
@kmiyashiro it looks like its only included one time. It looks like theres some checks inside ember-data for this case. https://github.com/emberjs/data/blob/f261ca8902219244994e79ba4785607d845d53ce/index.js#L68

workmanw:
@slindberg et al. I would be willing to try to a PR for this. Steven it sounds like you have some domain knowledge here. If you can give me a little bit of direction, I'm happy to take it and try to run with it.

dfreeman:
@slindberg @workmanw I'm also happy to put in some cycles if it helps move things forward here.

It looks like the build is currently pretty far off the ranch from the out-of-the-box addon setup (particularly with stubbing out the DS modules and the two-part npm/bower distribution) – is that in order to support producing a standalone artifact?

slindberg:
@dfreeman It's mostly a hold-over from pre-addon days. It was the pattern a lot of early addons adopted, so I went with it. I will be very happy when the build artifacts are no longer a part of the repo (or the release process), and updating MF takes place only in `package.json`.

[...]

PR #187, opened 2016-03-22
WIP - Convert to Ember-CLI Addon 

workmanw:
Originally written by @jakesjews in #186: 

> There is still a bit of cleanup to do but I wanted to get this out in the open. So far this seems to be working when I pull it into my company's applications. The tests are all passing locally but most of them are making deprecated calls to application.\_\_container\_\_. I've only really used mocha for testing ember apps so if anyone who's better at QUnit has some suggestions that would be very helpful.

TODO:
- [x] Replace deprecated calls to application.\_\_container\_\_ in the tests [@workmanw]
- [x] Restore `expectNoDeprecation` [@workmanw]
- [x] [@workmanw] code review, check merged commits, test in Batterii.
- [x] Fix debug version number display
- [x] Updating README.md [@workmanw]~~ See PR #189
- [x] Do we need the `ember-new-computed` polyfill? [@workmanw]
- [x] Update CHANGELOG.md~~ (will update when releasing)
- [x] Pair down package.json [@jakesjews]
- [x] Move bower dependencies to devDependencies [@jakesjews]
- [x] Fix `npm test`. I think ember-try just needs configured w/ supported versions. See [ember-try](https://github.com/kategengler/ember-try#config) [@workmanw].
- [x] Import model fragments before initializing the store. [@jakesjews]


workmanw:
@jakesjews @slindberg I believe that `ember-new-computed`  needs to be marked by `dependencies` rather than `devDependencies`. When I loaded up this in my app, I got an error about how the package could not be imported (I do not include it in my package.json). Adding it to `dependencies` fixed that.

I'm wonder if we even need that though. Since we're only targeting more modern versions of ember with this release, I think we can just remove the polyfill. Would you guys agree? If so, I can take care of that.

jakesjews:
@workmanw I threw that in after trying it in our app so it probably does need to be moved. I think its ok to remove it since we're targeting newer ember versions.

workmanw:
Submitted PR for README.md docs upgrade. #189.

bower.json
```diff
@@ -1,30 +1,10 @@
...
-  "devDependencies": {
-    "jquery": "^2.0.0",
-    "ember": "^2.1.0",
-    "loader.js": "~3.2.1",
-    "ember-dev": "emberjs/ember-dev#bcfb9c3487ec2fd58b932394a15ce16fd9cf7eed"
+    "ember": "~2.4.3",
+    "ember-cli-shims": "0.1.0",
+    "pretender": "^1.0.0",
+    "ember-cli-test-loader": "0.2.2",
+    "ember-qunit-notifications": "0.1.0"
```
slindberg: Is there a reason these are under `dependencies` and not `devDependencies`?

jakesjews:
No reason. I can fix it tomorrow.

index.js
```diff
@@ -0,0 +1,6 @@
+/* jshint node: true */
+'use strict';
+
+module.exports = {
+  name: 'model-fragments'
```
slindberg: Crap, does this have to be `model-fragments` to match the module name? I regret not using the full `ember-data-model-fragments` already :grimacing:

package.json
```diff
@@ -21,33 +24,38 @@
...
+    "ember-cli-uglify": "^1.2.0",
+    "ember-data": "^2.4.0",
+    "ember-dev": "emberjs/ember-dev",
+    "ember-disable-prototype-extensions": "^1.1.0",
+    "ember-disable-proxy-controllers": "^1.0.1",
+    "ember-export-application-global": "^1.0.4",
+    "ember-load-initializers": "0.5.1",
+    "ember-resolver": "^2.0.3",
+    "ember-try": "^0.2.2",
+    "loader.js": "^4.0.0"
```
slindberg: @jakesjews I assume these packages were the result of running `ember init` in the root? It looks like a good chunk of them are unnecessary for our development, could you pare it down to only the ones we need?

index.js
jakesjews: yeah unfortunately

jakesjews:
@slindberg @workmanw do you guys feel strongly about replacing the use of moduleForAcceptance in the unit tests. I'm thinking about removing it from the todo list if neither of you are concerned about it.

workmanw:
@jakesjews I don't have strong feelings at all. Honestly, I'm not sure what the viable alternative is. In our app, we have unit tests that are a maintenance nightmare because we have to build an application / owner and seed with basic DS classes to even use the store. This was an old pattern, so I'm sure there has to be a newer, better approach.

If there is a viable option for unit tests, perhaps it's worth a quick look. But as far as `moduleForAcceptance` is concerned, it makes things easier and it has no negative impact on the test suite AFAIK. ¯\\\_(ツ)\_/¯

jakesjews:
@workmanw We actually have some of the same issues in our app so I think it must be a pretty common problem.

I tried for awhile to find an alternative to moduleForAcceptance but didn't come up with anything. I'm just going to punt it off the list for now.

workmanw:
> I tried for awhile to find an alternative to moduleForAcceptance but didn't come up with anything. I'm just going to punt it off the list for now.

Agreed. We can always revisit this later if need be.

index.js
slindberg: I have a feeling this will interfere with installing the module itself. Need to test it out with `npm link`.

jakesjews: It should be ok. I've run it in my applications and I think @workmanw has tried it in his.

workmanw: Sorry, I don't follow the concern here. @slindberg are you saying for consistency reasons or do you think that this will cause trouble with ember-cli? Are there any other packages that have a different name like this?

In terms of what the export name should be, I don't have a strong opinion either way. I do think the name of addon should be "ember-data-model-fragments". So then the question becomes should the module name match. 

@jakesjews FWIW I did use this package in my app without any issue ... but I didn't do `npm install ...` or `ember install ...`. I just included it in my package.json, eg: `"ember-data-model-fragments": "lytics/ember-data-model-fragments#addonify`. I'm not sure if that makes a difference.

jakesjews: @workmanw good call on needing to check `ember install` I haven't actually tried that. As for the namespace I'm in favor of keeping it the same for backwards compatibility. We could definitely revisit it in the future though.

jakesjews:
@workmanw @slindberg the build with ember beta seems to be really flaky. Should we add beta to allowed failures in ember-try. I could see it discouraging PRs.

slindberg:
It is a shame that an application instance needs to be created just to run unit tests. The test suite is noticeably slower, but as long as it doesn't impede development it's not a big issue. I suppose it's better than [the custom registry voodoo](https://github.com/lytics/ember-data-model-fragments/blob/v2.1.2/tests/ember-setup.js) that is happening now too.

@workmanw if you end up finding a solution to unit tests that don't require `startApp`, let me know.

@jakesjews did you restart the travis build? I can look into why it fails; it would be really nice to get some advanced notice of beta incompatibilities.

jakesjews:
@slindberg yeah sorry. I wanted to confirm it was actually flaky and not something else.

workmanw:
> @workmanw @slindberg the build with ember beta seems to be really flaky. Should we add beta to allowed failures in ember-try. I could see it discouraging PRs.

I think it already is configured this way, see: https://github.com/lytics/ember-data-model-fragments/pull/187/files#diff-30

I'm still trying to learn the details of how ember-try works. Right now it seems to be running all the tests against one version:

```
------ RESULTS ------

Scenario default: SUCCESS
Command run: ember test
Scenario ember-2-3: FAIL
Command run: ember test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ ~2.3.0             │ 2.4.3                        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘

Scenario ember-release: SUCCESS
Command run: ember test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ components/ember#… │ 2.4.3                        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘

Scenario ember-beta: SUCCESS
Command run: ember test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ components/ember#… │ 2.4.3                        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘

Scenario ember-canary: SUCCESS
Command run: ember test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ components/ember#… │ 2.4.3                        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘


1 scenarios failed
4 scenarios succeeded
5 scenarios run
npm ERR! Test failed.  See above for more details.
```

I have all afternoon to work on this. Should have it figured out by tonight.

slindberg:
Oh weird. I've never used ember-try either, but I'm glad we are now :+1:

> EDIT: I think we also want to be testing lockstep versions of ember and ember-data as part of this. I'll see if I can figure out how to do that.

How do you mean? Since master is only intended to be compatible with ember-release or ember-beta, I don't think testing against previous versions will work.

workmanw:
> How do you mean? Since master is only intended to be compatible with ember-release or ember-beta, I don't think testing against previous versions will work.

My goal was to test the following:
 * ember and ember-data 2.3 - oldest version of we support
 * ember and ember-data release - Obviously the most recent version
 * ember and ember-data beta - With `allowedToFail: true` 
 * (maybe?) ember and emebr-data canary - with `allowedToFail: true`

The thought was to at least cover the boundaries of ember version range we support. So right now just 2.3 and 2.4. But after the next release it would be to test 2.3 and 2.5, then 2.3 and 2.6. Does that make sense?

slindberg:
Those bullets look good, we just need to make sure that any PRs that introduce incompatible changes bump the oldest version bullet (which will probably make it the same as `ember-release`).

Have either of you tried going through the process of `ember install ember-data-model-fragments` with an `npm link`ed package? I have a feeling the `model-fragments`/`ember-data-model-fragments` mismatch is going to make it fail.

jakesjews:
@slindberg I think I got it working with npm link. The steps I tried were
`npm link` in the model fragments addonify branch directory
`ember new test-app`
added `"ember-data-model-fragments": "*"` to package.json in test-app
```
npm link ember-data-model-fragments
ember s
```

When I ran `ember install ember-data-model-fragments` it seemed to still pull the regular version despite the npm link so I think I must be doing something wrong.

workmanw:
> Those bullets look good, we just need to make sure that any PRs that introduce incompatible changes bump the oldest version bullet (which will probably make it the same as ember-release).

Yeap! I completely agree. This should help us identify incompatibilities easily.


-------

I almost have this working. I think there is a small bug with ember-try and npm I'm going to isolate. I have a workaround that isn't too bad. But this is a really awesome tool. How it works is you basically define "scenarios" such as "ember/ember-data 2.3", each scenario has `package.json` and `bower.json` overrides. So it merges those into your files and does `npm install` and `bower install`. So you could create scenarios for anything!


```
------ RESULTS ------

Scenario default: SUCCESS
Command run: ember test
Scenario ember-2-3: SUCCESS
Command run: npm run install-and-ember-test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember-data         │ 2.3.3              │ 2.3.3                        │ npm      │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ 2.3.1              │ 2.3.1                        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘

Scenario ember-release: SUCCESS
Command run: npm run install-and-ember-test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember-data         │ emberjs/data#rele… │ 2.4.3                        │ npm      │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ components/ember#… │ 2.4.3                        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘

Scenario ember-beta: FAIL (Allowed)
Command run: npm run install-and-ember-test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember-data         │ emberjs/data#beta  │ 2.5.0-beta.3                 │ npm      │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ components/ember#… │ 2.5.0-beta.3+e596e13d        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘

Scenario ember-canary: FAIL (Allowed)
Command run: npm run install-and-ember-test
┌────────────────────┬────────────────────┬──────────────────────────────┬──────────┐
│ Dependency         │ Expected           │ Used                         │ Type     │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember-data         │ emberjs/data#mast… │ 2.6.0-canary                 │ npm      │
├────────────────────┼────────────────────┼──────────────────────────────┼──────────┤
│ ember              │ components/ember#… │ 2.6.0-canary+9c645581        │ bower    │
└────────────────────┴────────────────────┴──────────────────────────────┴──────────┘


2 scenarios failed (2 allowed)
3 scenarios succeeded
5 scenarios run
```

workmanw:
I can't seem to figure out how to use get the `npm link` trick to work either with ember install. Happy to try it out, just need a bit of help.

slindberg:
Ah sorry, didn't realize `npm link`ing a module doesn't make it installable, must have always done it manually and not realized it. I just want to make sure everything is copacetic once it's installed, i.e. generators work, `model-fragments` import is recognized, etc.

workmanw:
@slindberg @jakesjews All the checklist items are complete. Can you guys think of any outstanding tasks here? 

I know Steven expressed some concerns about the module name and the install process. One thing that occurred to me was we could cut a beta release of 2.3. We could try it in our apps and new apps. Other people from #174 may also be willing to try it. If we did decide to change the module name, it should be as simple as find and replace for beta users of 2.3.

jakesjews:
I think a beta release is the way to go. I can pull the beta into our application and run it through QA this week to see if anything shakes out.

slindberg:
I created a test app this weekend and used `npm link` to manually install the package. I added a model with a fragment attribute, and an application route that creates a record in its `model` hook. When the app is loaded, I get the error:

```
Error while processing route: index store.createFragment is not a function TypeError: store.createFragment is not a function
```

It looks like the store instance is getting created before MF can monkey patch it, which sort of makes sense since it's probably lazy-loaded now via the first module import. I think maybe the only way to fix this is to do a `import _ from 'model-fragments'` in an initializer before the store can get instantiated.

@workmanw I don't think it's necessary to release a 2.3 beta, since you can test it using `npm link` in your app. If it's a pain for some reason though, I definitely can.

slindberg:
@jakesjews I also noticed that the version stamping isn't working because it's trying to replace an identifier that is in the license (`config/license.js`), which doesn't get prepended anymore. Can you revert that build step too?

jakesjews:
@slindberg my bad I could have sworn I fixed that. Are you referring to the version number thats shows in the console?

jakesjews:
@slindberg my broccoli-fu is a little lacking. Do you know how I can replace the usage of outputFile in prependLicense? It looks like I just have an app tree reference now.

jakesjews:
@slindberg I added an import of model-fragments in the initializer so it should run before the store gets initialized. I'm still having some trouble getting the license stuff to work but hopefully I'll figure it out soon.

jakesjews:
I think I found a way to get the version number working again but there are some drawbacks. I can use the treeForAddon hook in index.js but it looks like that runs whenever a user builds as opposed to before publishing the addon. Would some sort of publish hook using ember-cli-release be the way to go? We might need to change the version placeholder a bit since the actual source files will be modified now instead of a dist file. There is probably a better way to do it so I'm open to other options.

slindberg:
> I can use the treeForAddon hook in index.js but it looks like that runs whenever a user builds as opposed to before publishing the addon. Would some sort of publish hook using ember-cli-release be the way to go?

Since we're moving to addon-only distribution, there are no build artifacts in the release, so no need to worry about versioning before publish as long as the `package.json` version is correct. I don't think it's a big deal if the license prepending/version replacement happens at build time since it's adds insignificant overhead. This way, it's also guaranteed that the version in the built source is always correct (whatever is in `package.json`). If at some point it's necessary to create a stand-alone build, version replacement will happen as part of the build like normal.

app/initializers/model-fragments.js
```diff
@@ -1,4 +1,6 @@
-import _ from 'model-fragments';
+// Import the fragment base class to ensure monkey-patching side effects are
+// applied before any store instances are created :(
+import _ from 'model-fragments/fragment';
```
workmanw: You actually don't even need to do the `_ from` if you don't want to. `import 'model-fragments/fragment';` is valid. We actually do this once on the test-helper already. https://github.com/lytics/ember-data-model-fragments/blob/addonify/tests/test-helper.js#L1

slindberg: Haha derp, I knew that I swear :wink: 

I actually realized that I need to revert the `/fragments` portion, because the index file is where MF gets registered with `Ember.libraries`.

slindberg:
@jakesjews I have a couple spare hours, so I'm going to look into the license appending. I also realized that ember data doesn't even append their license to the addon build, so maybe it's not even worth it. Will update.

slindberg:
I decided not to worry about adding the license to the build, mainly since all comments seem to be stripped out of production builds anyway.

The only thing left that gives me pause is that I realized ED sill supports globals if it is specified in `bower.json` (look for `_forceBowerUsage` [here](https://github.com/emberjs/data/blob/master/index.js)). I really want to just ignore it and force users of 2.3.x to use addons, but I don't have a good sense for how many people are not using ember-cli but are running release versions of Ember/ED. What do you guys think?

Also, @workmanw were you ever able to test this branch in your app?

workmanw:
@slindberg Yeap! We tested this branch without issue. All of our integration tests passed and everything worked. I also ran into another model-fragments user at emberconf who said he was testing this branch and it was working for him. Talk about a coincidence. :+1:

jakesjews:
@slindberg I think it's ok to remove the global build. I can't think of any other addons that support globals.

slindberg:
> I think there is an implicit assumption that you need to use ember-cli to work with addons

True, but this project has been a globals library up to this point. Either way, I think we just go with it until someone asks. I think it's time to merge this puppy :tada: 

Thank you so much for all of your work on this project @jakesjews & @workmanw! Especially for the painful porting of tests to the addon infrastructure. Go team!

mattmcginnis:
Should this get a version bump?

slindberg:
It will, don't worry :wink:

PR #189, opened 2016-03-23
First pass at updating the README.md to match the new addonify patterns.

workmanw:
I'm submitting this as a PR to the addonify branch so that @slindberg and @jakesjews can review it. Here is a link to the [rendered view](https://github.com/workmanw/ember-data-model-fragments/blob/d15613cf9d62c30e604dc08d5031adeddc7e975a/README.md).

There are two things I specifically wanted to get feedback on. 

1. Since I've upgraded the docs to use the newish ES6 model pattern (rather than globals) the examples tend to have more files. See [here](https://github.com/workmanw/ember-data-model-fragments/blob/d15613cf9d62c30e604dc08d5031adeddc7e975a/README.md#example). I added the name of the model as a path based code comment at the top of each snippet. This is what many addons do, include the official ember model guides. My question is, is that clear enough for someone reading it? If not, I think a good alternative would be to make a variable, we do this in our app. Eg.

    ```javascript
    // app/models/person.js
    import Model from 'ember-data/model';
    import {
      fragment,
      fragmentArray,
      array
    } from 'model-fragments/attributes';

    let PersonModel = Model.extend({
      name      : fragment('name'),
      addresses : fragmentArray('address'),
      titles    : array()
    });

    export default PersonModel;
    ```

2. `fragment` and `Fragment` look kind of ambiguous in the documentation. If you didn't catch the difference, the former is the attribute marco while the later is the model class. Just looking at it, it seems ripe for confusion. Perhaps I should change `fragment` to `fragmentAttr`? But then what about `fragmentArray` and `array`, would you call them `fragmentArrayAttr` and `arrayAttr` respectfully? Thoughts?

TODO:
- [x] We probably don't need to meantion `ember build` any more.
- [x] Document how to use `testem` and `ember try:testall`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,9 +426,9 @@ import attr from 'ember-data/attr';
 import { fragmentArray } from 'model-fragments/attributes';
 
 export default Model.extend({
-  name: attr("string"),
-  city: attr("string"),
-  animals: fragmentArray("animal", { polymorphic: true, typeKey: '$type' }),
+  name: attr('string'),
+  city: attr('string'),
+  animals: fragmentArray('animal', { polymorphic: true, typeKey: '$type' }),
 });
 ```
 
@@ -438,27 +438,27 @@ import Fragment from 'model-fragments/fragment';
 import attr from 'ember-data/attr';
 
 App.Animal = Fragment.extend({
-  name: attr("string"),
+  name: attr('string'),
 });
 ```
 
 ```javascript
 // app/models/elephant.js
-import Animal from "./Animal";
+import Animal from './Animal';
 import attr from 'ember-data/attr';
 
 export default Animal.extend({
-  trunkLength: attr("number"),
+  trunkLength: attr('number'),
 });
 ```
 
 ```javascript
 // app/models/lion.js
-import Animal from "./Animal";
+import Animal from './Animal';
 import attr from 'ember-data/attr';
 
 export default Animal.extend({
-  hasManes: attr("boolean"),
+  hasManes: attr('boolean'),
 });
 ```
 
@@ -518,7 +518,7 @@ export default JSONSerializer.extend({
 });
 ```
 
-```javscript
+```javascript
 // app/serializers/elephant.js
 import AnimalSerializer from './animal';
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Use the following table to decide which version of this project to use with your
 | >= v1.0.0-beta.15 <= v1.0.0-beta.18 | v0.3.3 |
 | >= v1.13.x < v2.0.0 | v1.13.x |
 | >= v2.0.x < v2.1.0 | v2.0.x |
-| >= v2.1.x | v2.1.x |
+| >= v2.1.x < v2.3.x | v2.1.x |
 | >= v2.3.x | v2.3.x |
 
 #### Notes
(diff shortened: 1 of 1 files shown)


jakesjews:
@workmanw the readme looks good!

As for your two questions

- I think it is preferable to follow whatever the ember documentation convention is so I think what you have is good
-  It definitely is a bit confusing to have fragment and Fragment mean different things but for the sake of backwards compatibility I'm in favor of keeping it the same.

workmanw:
@jakesjews Thanks!

>It definitely is a bit confusing to have fragment and Fragment mean different things but for the sake of backwards compatibility I'm in favor of keeping it the same.

Yea this is exactly how I feel. In thinking about it more, really we kind of has this problem with the namespace (IE `MF.fragment` vs `MF.Fragment`) and now it's really just a documentation thing. Users could choose to do `import { fragment as fragmentAttr } from 'model-fragments/attributes';` if they wanted to. Yay ES6!

I'll leave this open for a few days to give @slindberg a chance to review. I'm also trying to get another dev I work with to give it a once over, but he's strapped for time at the moment.

README.md
```diff
@@ -22,12 +22,14 @@ Use the following table to decide which version of this project to use with your
 | >= v1.13.x < v2.0.0 | v1.13.x |
 | >= v2.0.x < v2.1.0 | v2.0.x |
 | >= v2.1.x | v2.1.x |
```
slindberg: This should change to `| >= v2.1.x < v2.3.x | v2.1.x |`

README.md
```diff
@@ -22,12 +22,14 @@ Use the following table to decide which version of this project to use with your
...
+| >= v2.3.x | v2.3.x |
 
 #### Notes
 
 - Ember Data v1.0.0-beta.12 introduced a bug that makes it incompatible with any version of this project.
 - Ember Data v1.0.0-beta.15 introduced a breaking change to the serializer API with [Snapshots](https://github.com/emberjs/data/pull/2623). Since this affected fragment serialization as well, support for it was added in v0.3.0. See the [serializing](#serializing) section below for more information.
 - Ember Data v1.0.0-beta.19 refactored a large number of internal APIs this project relied on and is not officially supported. Compatibility was added in v0.4.0 and targeted at Ember Data v1.13.x.
+- Ember Data 2.3 converted to a full Ember CLI addon. Removing the global `DS` namespace and switching to an import module strategy. More: [Ember Data 2.3 Released](http://emberjs.com/blog/2016/01/12/ember-data-2-3-released.html).
```
slindberg: Should probably mention getting rid of the `MF` global too.

README.md
workmanw: @slindberg Good catch!

README.md
workmanw: @slindberg I'm on it! Thank you!

workmanw:
@slindberg @jakesjews I did have a colleague review these changes. His feedback was to make all the the code snippet use single-quotes. Other than that, he was good on it. If there are no objections, I'll likely merge this into the addonify branch shortly.

slindberg:
I think the examples are plenty clear since you can read the import directly, I'm actually more worried about removing the `DS` and `MF` namespaces from the text, since it's not obvious where they are coming from anymore. Without keeping the namespaces though, I don't what the answer is. It'd be a little weird to refer to `ember-data/Transform` or `model-fragments/Fragment`...

As for `fragment` vs. `Fragment` I agree it's not ideal, but changing either would break semver, so they are pretty locked in. :shrug:

Issue #191, opened 2016-04-01
Deprecation warning "Calling Ember.get without a target object" on 1.13.3


FredUK:
Hi, first of all thanks a lot for this great addon.

I've been for the past few days trying to update my Ember app to 1.13.13 with Ember data 1.13.15 with your latest Data Model Fragments 1,13,3 before I even try update to v2 and noticed a deprecation warning I can't seem to fix and according to the call stack the last call before the warning, comes from [fragment.js#L129](https://github.com/lytics/ember-data-model-fragments/blob/f7ac7fce03701587e56473cca63a85b8ca640267/addon/fragment.js#L129)

![image](https://cloud.githubusercontent.com/assets/11350311/14212671/e12f5db8-f82b-11e5-8056-88bed1857481.png)

Can you confirm and perhaps shed any light why this is/might be happening?

Thanks a lot.
Cheers, Fred

slindberg:
It looks like what's happening is `toString` is getting called on a fragment that hasn't been added to a record yet (i.e. created with `store.createFragment`, but never added to a record with `record.set('prop', fragment')`). Usually this happens when debugging, since records/fragments rarely get coerced into strings under normal circumstances; are you `console.log`ing the record in that call stack?

We definitely should be checking whether the fragment has an owner record before calling `Ember.get` on it, so this is a bug. In v2 calling `Ember.get(null)` throws an exception, so this code path will definitely break your app. I would suggest avoiding calling `toString` on fragments until it's fixed. Thanks for reporting.

slindberg:
Fixed by #194 and included in v2.3.1

PR #194, opened 2016-04-12
ember 2.5 compatibility

jakesjews:
Replace deprecated use of Ember.merge with Ember.assign.

Right now an assertion is being triggered in ember-data since Ember.merge is being used and this is causing a null exception in toString() since it's called before the owner is set. It seems like there could be other cases where toString could be called before a parent is assigned so this might be a useful fix.

### DIFF
--- a/addon/fragment.js
+++ b/addon/fragment.js
@@ -126,7 +126,12 @@ var Fragment = Model.extend(Ember.Comparable, Ember.Copyable, {
   },
 
   toStringExtension: function() {
-    return 'owner(' + get(internalModelFor(this)._owner, 'id') + ')';
+    let owner = internalModelFor(this)._owner;
+    if (owner) {
+      return 'owner(' + get(owner, 'id') + ')';
+    } else {
+      return '';
+    }
   }
 }).reopenClass({
   fragmentOwnerProperties: Ember.computed(function() {
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "ember-data-model-fragments",
   "devDependencies": {
-    "ember": "~2.4.3",
+    "ember": "~2.5.0",
     "ember-cli-shims": "0.1.1",
     "pretender": "^1.0.0",
     "ember-cli-test-loader": "0.2.2",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "devDependencies": {
     "broccoli-asset-rev": "^2.2.0",
-    "ember-cli": "2.4.2",
+    "ember-cli": "2.4.3",
     "ember-cli-dependency-checker": "^1.2.0",
     "ember-cli-htmlbars": "^1.0.1",
     "ember-cli-inject-live-reload": "^1.3.1",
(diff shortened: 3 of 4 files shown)


workmanw:
👏 👏 👏  I'll make time this afternoon to test this in our app, but just code reviewing it seems straightforward.

addon/fragment.js
```diff
@@ -126,7 +126,12 @@ var Fragment = Model.extend(Ember.Comparable, Ember.Copyable, {
   },
 
   toStringExtension: function() {
-    return 'owner(' + get(internalModelFor(this)._owner, 'id') + ')';
+    let owner = internalModelFor(this)._owner;
+    if (owner) {
+      return 'owner(' + get(owner, 'id') + ')';
+    } else {
+      return '';
+    }
```
workmanw: Was this this a change that resulted from ember/ember-data 2.5 or was it just something to fix while you were in there?

workmanw:
I did test this in our app, and as expected it worked just fine. I'm ready to merge this once I hear back about that code comment I left. 

Should we release this as a patch release? IE `2.3.1`? The `Ember.merge` change is just for the test suite, so really it would just be the `toStringExtension` change.

addon/fragment.js
jakesjews: I actually found the problem with the combination of Ember-Data 2.4 and Ember 2.5. It was triggering a deprecation from Ember-Data using Ember.merge which was causing errors in this function. The problem doesn't seem to exist with Ember-Data 2.5 but I think this is good to prevent crashes.

workmanw:
@jakesjews Do you think this we should cut `2.3.1`? If so, I can submit a PR to bump the version and we can tag it up. We'd still need @slindberg to publish to npm, but at least that would save him a few steps.

jakesjews:
@workmanw yeah that would be awesome

addon/fragment.js
slindberg: Sweet, this should fix #191

bower.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ember-data-model-fragments",
   "devDependencies": {
-    "ember": "~2.4.3",
+    "ember": "~2.5.0",
```
slindberg: I think this means we need to create an `ember-try` case for 2.4.x

slindberg:
Thanks @jakesjews! It looks like the `Ember.assign` fix was only a testing fix? Were you seeing a deprecation warning in your app somehow?

Also, just to make sure, this is fully compatible with 2.3.x, correct? If so, I agree this belongs in a `2.3.1` release. Release management is done using `ember-cli-release`, so there's no advantage to creating the tag ahead of time, especially since the tagged release should include a changelog commit (in fact it is probably more of a hassle).

bower.json
workmanw: @slindberg Yeap, I think that's a good idea. I'll submit a PR for this right now.

jakesjews:
@slindberg There was an error being thrown in our app from some complicated combination of a deprecation from somewhere else in the app calling toString on a fragment. It looks like the deprecation might not be as big of an issue soon since they're un-deprecating Ember.merge https://github.com/emberjs/ember.js/commit/1fb2a2ef381588a84f8e55e306a94c50016320b2

PR #212, opened 2016-08-31
Fix `internalModel is null` error in `toStringExtension`

cesarizu:
When using mixed adapters sometimes `internalModel` is null and an error is thrown. This PR checks to see if internalModel is set so that the `toStringExtension` doesn't fail.

### DIFF
--- a/addon/fragment.js
+++ b/addon/fragment.js
@@ -126,7 +126,8 @@ var Fragment = Model.extend(Ember.Comparable, Ember.Copyable, {
   },
 
   toStringExtension: function() {
-    let owner = internalModelFor(this)._owner;
+    let internalModel = internalModelFor(this);
+    let owner = internalModel && internalModel._owner;
     if (owner) {
       return 'owner(' + get(owner, 'id') + ')';
     } else {
@@ -174,7 +175,7 @@ export function internalModelFor(record) {
 
   // Ensure the internal model has a fragments hash, since we can't override the
   // constructor function anymore
-  if (!internalModel._fragments) {
+  if (internalModel && !internalModel._fragments) {
     internalModel._fragments = create(null);
   }
 


workmanw:
@cesarizu This looks good to me. Do you have any details on why `internalModel` is sometimes null? Would it be possible to write a test for this?

cesarizu:
On my side it only happens on models that use mixed adapters. That is, models that use one adapter and that have relationships to other models that use a different adapter. On my case I'm using a jsonapi adapter and a local storage adapter. See https://github.com/emberjs/data/pull/4387 for a related error.

It seems to be a bug in ember-data and I've been trying to figure out a way to test this but haven't figured out how to do that. Maybe if you have any idea on how to do that I could give it a try.

cesarizu:
Yes, with this PR it does work, without it I get lots of "internal model is null" errors. I consider this PR a workaround, since I'm pretty sure the error is in ember data. I'll try to setup a simple app that reproduces this error first to try to isolate the exact problem.

workmanw:
@cesarizu I would greatly appreciate that. A simple ember-twiddle would suffice too.

ToMoCoop:
I'm using your GovSciences:internal-model branch at the moment and so far it seems successful and resolves the errors.
Whether this counts for anything, I don't know!

jakesjews:
@workmanw I'm able to replicate the issue at https://github.com/jakesjews/ModelFragmentObserver. I think we should merge this since I'm not quite sure of another way to work around the issue. Whats happening is when we call [modelFor at the beginning of createFragment](https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-metal/lib/events.js#L85) it is triggering an [assertion test inside ember](https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-metal/lib/events.js#L85) which calls toString on the fragment before it is created so internalModel is null. It doesn't really look like theres anything to key off for toString instead of the parent models id so I think a null check is at least a good stop gap for now.

workmanw:
@jakesjews Sorry I just saw this message. Thanks for the reproduction, I definitely appreciate it. I'll review this after lunch.

workmanw:
Alright, sorry for the delay. I'm 👍  on merging this.

jakesjews:
Awesome thanks for taking a look

PR #214, opened 2016-10-06
updates polymorphisme example

basz:
The example did not work for me without this patch. (tried with Ember(Data) 2.7 and 2.8).

Perhaps I did it [incorrect](https://github.com/lytics/ember-data-model-fragments/issues/213) but I found that the following method could never work as the `data` argument did not contain the typeKey's field.

https://github.com/lytics/ember-data-model-fragments/blob/master/addon/fragment.js#L160

```
export function getActualFragmentType(declaredType, options, data) {
  console.log(declaredType, options, data); // would log: normalize Object {name: "Leonard", hasManes: true}
  if (!options.polymorphic || !data) {
    return declaredType;
  }

  var typeKey = options.typeKey || 'type';
  var actualType = data[typeKey];

  return actualType || declaredType;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,7 +467,7 @@ import Fragment from 'model-fragments/fragment';
 import attr from 'ember-data/attr';
 
 App.Animal = Fragment.extend({
-  $type: attr('string),
+  $type: attr('string'),
   name: attr('string'),
 });
 ```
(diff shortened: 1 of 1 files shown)


jakesjews:
I've run into something similar myself before thanks for the PR. @workmanw this LGTM do you think this is good to go?

README.md
```diff
@@ -467,6 +467,7 @@ import Fragment from 'model-fragments/fragment';
 import attr from 'ember-data/attr';
 
 App.Animal = Fragment.extend({
+  $type: attr('string),
```
workmanw: @basz If you close out this string, I'm :+1:. Thank you!!

basz: done

workmanw: :+1: Thanks!

PR #216, opened 2016-10-21
update to ember-cli 2.9.1

jakesjews:


jakesjews:
@workmanw I think we might be due for a release after this PR

workmanw:
@jakesjews Sorry I've been late on responding. I'm going to review this first thing tomorrow morning. I'm very much 👍 on switching ember-try to use LTS releases for backwards compat. I think that's a smart move. Eyeballing, everything looks in order here. I'll clone and confirm tomorrow.

jakesjews:
For some reason CI is failing on canary even though it's set to allow failures so I gotta look into that too.

jakesjews:
I ended up removing the canary ember-try scenario for now. It looks like its hitting some sort of infinite loop which is causing CI to fail.

workmanw:
Really sorry about the delay this week. Thanks for keeping up with this. I'm very much :+1: on checking the LTS versions, I think that's a great idea.

jakesjews:
no problem there isn't really a rush on this

PR #225, opened 2016-12-06
update to ember-cli 2.10.0

jakesjews:


jakesjews:
@slindberg @workmanw I'm thinking we should release after this. @slindberg when would you be available to go through the release process?

jakesjews:
I'm not sure what to do about the beta fails with regards to https://github.com/lytics/ember-data-model-fragments/issues/224

workmanw:
Hey @jakesjews -- I believe I have things squared away for ember-data-beta (2.11.x). Unfortunately, there will be a non-backwards compatible change and we'll have to bump our incompatibility.

For that reason I'd like to do a few things here.
1. I'd like to go ahead and merge this. 
2. Cut a new tag release after that for a patch update (v2.3.3). 
3. After that, we'll get my PR submitted / reviewed / merged. Then we'll bump this version to 2.11.

Does that all make sense and sound right?

jakesjews:
@workmanw yeah that sounds good. Thanks for taking the lead on the fixes!

PR #226, opened 2016-12-12
Use the `ember-data` initializer as a target instead of the soon to b…

bmac:
…e deprecated `store` initializer

Hello, The `store` initializer target currently is a [empty initializer](https://github.com/emberjs/data/blob/master/app/initializers/store.js#L10-L14) in Ember Data. I have an [RFC](https://github.com/emberjs/rfcs/pull/181) and a [PR](https://github.com/emberjs/data/pull/4657) to deprecate it a future Ember Data release. I did a search and found your addon is using the `store` initializer  target and I have opened this pr for you to use the `'ember-data'` initializer instead which will continue to be supported after the empty initializers are deprecated.

### DIFF
--- a/app/initializers/model-fragments.js
+++ b/app/initializers/model-fragments.js
@@ -7,7 +7,7 @@ import ArrayTransform from 'model-fragments/transforms/array';
 
 export default {
   name: "fragmentTransform",
-  before: "store",
+  before: "ember-data",
 
   initialize: function(application) {
     application.register('transform:fragment', FragmentTransform);


jakesjews:
Will this be backwards compatible with previous versions of ember-data?

bmac:
Yes the `ember-data` initializer has existed since Ember Data `1.0.0-beta.8`.

jakesjews:
Awesome LGTM then. @workmanw I think we should merge this since stable ember-data is passing

workmanw:
Thanks @bmac! -- @jakesjews Agreed. Merging. The travis-ci failure is from beta, which I'm working to fix as we speak.

PR #227, opened 2016-12-12
[WIP] Adding support for ember-data 2.11.

workmanw:
Adding support for Ember-Data 2.11. Fixing #224. Unfortunately this change is not going to be backwards compatible with `>= ember-data v2-10.x`.

**This is still WIP. DO NOT MERGE** 😄. My plan is to merge this after Ember-Data 2.11.0 ships.

------------

Just leaving some notes here so I can remember in 4-5 weeks 😜 .
TODOs -- Once Ember-Data `2.11.0` ships:

- [x] Restore the `ember-release` scenario in `.travis.yml`.
- [x] Update package.json to use Ember-Data-2.11 as default.
- [ ] Update CHANGELOG.md
  - [ ] Mention @runspired as co-author of solution.
  - [ ] Mention GavinJoyce for `store.isFragment` (c6834f8)
  - [ ] Mention cibernox for `Ember.K` (96203e34)
- [ ] Release v2.11.0. 
  - [ ] Update package.json to use Ember-2.11 as default via `ember-source`.
  - [ ] Update ember-try.js
  - [ ] Also update for ember-cli 2.11 release (**if timing works right**).

PR #235, opened 2017-01-15
Added `store.isFragment`

GavinJoyce:
[`ember-data-factory-guy`](https://github.com/danielspaniel/ember-data-factory-guy) currently relies on [`store.createFragment` raising an exception](https://github.com/danielspaniel/ember-data-factory-guy/blob/20ba57cafaa366b59e62fd9e1afef01d5d673c53/addon/model-definition.js#L41-L50) to detect if a model is a factory. This exception comes from an assertion which will be stripped in certain builds once https://github.com/ember-cli/rfcs/pull/50 is adopted.

This PR introduces a `store.isFragment` method which will allow ember-data-factory-guy to do something like:

```js
isModelAFragment() {
    if(FactoryGuy.store.isFragment) {
      return FactoryGuy.store.isFragment(this.modelName);
    }

    return false;
}
```

There are currently failing tests on master, and the same on this branch. https://github.com/lytics/ember-data-model-fragments/pull/227 seems to fix them and will likely land soon

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -46,7 +46,7 @@ Store.reopen({
   createFragment: function(modelName, props) {
     var type = this.modelFor(modelName);
 
-    Ember.assert("The '" + type + "' model must be a subclass of MF.Fragment", Fragment.detect(type));
+    Ember.assert("The '" + type + "' model must be a subclass of MF.Fragment", this.isFragment(type));
 
     var internalModel = new InternalModel(type, null, this, getOwner(this).container);
 
@@ -68,6 +68,19 @@ Store.reopen({
     fragment._isFragment = true;
 
     return fragment;
+  },
+
+  /**
+    Returns true if the modelName is a fragment, false if not
+
+    @method isFragment
+    @private
+    @param {String} the modelName to check if a fragment
+    @return {boolean}
+  */
+  isFragment(modelName) {
+    var type = this.modelFor(modelName);
+    return Fragment.detect(type);
   },
 
   /**
--- a/tests/unit/store_test.js
+++ b/tests/unit/store_test.js
@@ -38,6 +38,11 @@ test("attempting to create a fragment type that does not inherit from `MF.Fragme
   });
 });
 
+test("the store has an `isFragment` method", function(assert) {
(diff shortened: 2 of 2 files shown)


GavinJoyce:
I'm seeing some test failures (on master too). I'll take a look soon

workmanw:
@GavinJoyce Yeap, I know that there are more ember-data internal changes coming in `ember-data-2.12-beta` and canary. I was hoping to make time to investigate them in the coming weeks. 

I was waiting for `ember-2.12` to ship before releasing #227 because I thought it would only be a few days out. I'm not sure what the current status of that is. 

I am curious if you know why `ember-data-factory-guy` relied on this assertion rather than just using `detect` in the first place? Was it simpler to do that than try to get a hold of the `Fragment` base class? 

Either way, I'm okay with this addition, it seems straight forward enough to me. @jakesjews, any concerns with this?

jakesjews:
@workmanw definitely good with this. It has actually been difficult debugging on caught exceptions because of this in our tests so this will be nice.

workmanw:
I'll be sure to rebase #227 on top of this so it goes out. Thanks @GavinJoyce.

workmanw:
OH It looks like this is triggering a deprecation warning on master :(. I should have paid more attention to it, I'm sorry.

The problem is simple, code looks like this:
```javascript
createFragment: function(modelName, props) {
  var type = this.modelFor(modelName);

  Ember.assert("The '" + type + "' model must be a subclass of MF.Fragment" Fragment.detect(type));		 +    Ember.assert("The '" + type + "' model must be a subclass of MF.Fragment", this.isFragment(type));

  // ...
}, 

isFragment(modelName) {
  var type = this.modelFor(modelName);
  return Fragment.detect(type);
},
```

The problem is that when we call `this.isFragment(type)`, `isFragment turns around and calls `this.modelFor(modelName)` with the instance, rather than the name. I think the best solution is to just move the assert to the top of the function, and call `this.isFragment(modelName)`.

@jakesjews @GavinJoyce Sound right? If so, I'll make a quick PR.

GavinJoyce:
@workmanw Thanks, moving the assert up makes sense.

PR #236, opened 2017-01-18
Remove lookupSerializer deprecation

estshy:
On the latest stable build of ember data `lookupSerializer` is marked as deprecated and aliased towards `serializerFor` which makes infinite loop.

See https://github.com/emberjs/data/blob/v2.11.0/addon/-private/system/store.js#L2608

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -112,7 +112,7 @@ Store.reopen({
         '-default'
       ];
 
-      return this.lookupSerializer(modelName, fallbacks);
+      return this._super(modelName, fallbacks);
     }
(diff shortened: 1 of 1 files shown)


workmanw:
Hey @estshy . Thanks for the PR, I think this is covered by #227. Can you verify?

estshy:
@workmanw Yes, it seems to be covered by #227 as well.

workmanw:
I'm going to close this since #227 was landed. Apologies for the delay on resolving the issue. Thanks again for the PR.

Issue #232, opened 2017-01-05
Ember-Data 2.12 compat


workmanw:
We should track this and follow suit: 

- [x] Investigate: ember-data compat with factoryFor ([#4737](https://github.com/emberjs/data/issues/4737)) -- Fixed by #240.
- [x] Investigate: Deprecation "cannot pass class to store method" (#229) -- Fixed by #241.
- [x] Create point release.

workmanw:
I'll be taking a look into these issues in the next week or so. For reference, we should be able to follow the pattern @cibernox used here: https://github.com/jamesarosen/ember-i18n/pull/431 .

workmanw:
Alright, I think it's high time I shipped this point release. `ember-data-2.12.0-beta.3` has settled. I'm going to release `2.11.1` to be compatible.

PR #241, opened 2017-01-25
Fixing an Ember-Data 2.12 compatibility issue (#232).

workmanw:
Here is the upstream commit:  https://github.com/emberjs/data/commit/a5a8cd7714555655ae9fb7209b5b87c11aabd1cc 

That change was a perf improvement for ember-data, but we need to change the way we initialize our fragments to follow suite. This was pretty easy and we are able to be backwards compatible by detecting a property on the prototype. 

Let's also hold on merging this for the time being. I'd like to see the `factoryFor` PR land first. Then we can land this and all the tests should be passing.

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -19,6 +19,10 @@ var keys = Object.keys || Ember.keys;
 var create = Object.create || Ember.create;
 var getOwner = Ember.getOwner;
 
+var InternalModelPrototype = InternalModel.prototype;
+var internalModelExpectsModelName = InternalModelPrototype.hasOwnProperty('modelClass');
+
+
 /**
   @class Store
   @namespace DS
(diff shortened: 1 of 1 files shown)


workmanw:
WOOT! All tests are passing. Even beta and canary :) https://travis-ci.org/lytics/ember-data-model-fragments/builds/196463842

addon/ext.js
```diff
@@ -47,8 +47,15 @@ Store.reopen({
   createFragment: function(modelName, props) {
     Ember.assert("The '" + modelName + "' model must be a subclass of MF.Fragment", this.isFragment(modelName));
 
-    var type = this.modelFor(modelName);
-    var internalModel = new InternalModel(type, null, this, getOwner(this).container);
+    var internalModel;
+    if (InternalModelPrototype.hasOwnProperty('modelClass')) {
```
jakesjews: I'm not sure if it is an issue in v8 anymore but looks like hasOwnProperty has caused some [slowdown in the past](https://github.com/emberjs/data/pull/4313) with ember. Since InternalModelPrototype perhaps we could set the result of this check in a variable outside the function and reference that?

workmanw: Oh, I didn't know that! Thanks for catching this. Yea, I like your idea of just moving it outside the function. I'll make it so.

jakesjews:
@workmanw looks great!

workmanw:
@jakesjews Thanks!

Issue #245, opened 2017-01-31
Bug: "module name" should match the "package name"


workmanw:
The module name is `model-fragments`, while the package name is `ember-data-model-fragments`. This seems to cause a problem with ember-cli, it's unable to automatically detect `isDevelopingAddon` which resulted in this addon never being linted while the tests were running :(.

![image](https://cloud.githubusercontent.com/assets/825487/22476889/f2adcb08-e7b1-11e6-8377-0ecfe72c8593.png)

I spoke with @Turbo87 on slack. He recommended we change the module name to match the package name (which I agree with). His idea was to use a vendor/shim.js file to provide backwards compatibility. 

I'm planning to work on this.

workmanw:
Closed via #246

PR #246, opened 2017-01-31
Fixing issue #245. 

workmanw:
Unfortunately this requires use to change the module name of this addon from `model-fragments` to `ember-data-model-fragments`. Added a vendor shim to provide backwards compatibility.

See: #245. 

Two outstanding questions:

1. Are there any other implications here? _I can't think of any, it should just work._
2. Should we deprecate using `model-fragments` right away or wait a while?

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ import {
   fragment,
   fragmentArray,
   array
-} from 'model-fragments/attributes';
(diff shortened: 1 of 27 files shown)


jakesjews:
@workmanw This works in my work application so I think it should work fine. I vote in favor of deprecating now since it can probably be updated with a simple find and replace.

Issue #190, opened 2016-03-31
Observer causing Maximum call stack size exceeded with nested fragments


shidel-dev:
I am using
 ember data: 2.1.0
 ember: 2.1.0
 ember model fragments: 2.1.2

I have a model with a fragment that has a fragment. When using a computed alias or an observer, I get a Maximum call stack size exceeded after saving.

shidel-dev:
seems similar to https://github.com/lytics/ember-data-model-fragments/pull/60 and https://github.com/lytics/ember-data-model-fragments/pull/54

shidel-dev:
I have tried updating to ember 2.2, and ember data 2.2 (like what is present in the bower.json for ember-data-fragments@2.1.2) and this issue continues.

slindberg:
These issues have classically been impossible to diagnose without a failing test or app/twiddle that demonstrates the issue. Can you extract the bare minimum of logic from your app into a test case that reproduces the issue? Thanks for reporting.

workmanw:
Actually digging through our logs I found one case where this happened. It's not really clear exactly what the user did, but I suspect they may have done something like the following:

1) Create a new fragment on an existing record (in this case `fragment` not `fragmentArray`).
2) Save the record.
3) Update an attribute on the fragment *before* the record has finished saving.
4) When the adapter fulfills it's promise and the save is complete, that's when it looks like this error *might* have happened.

That's all just a very very rough guess. But I can confirm that we saw this issue with version 2.1 once in production.

DingoEatingFuzz:
While upgrading to Ember 2.2 I came across this in our app as well. I was able to reduce it in a demo app here: https://github.com/DingoEatingFuzz/mf-double-alias-bug

There is a model that has a fragment relationship, and that fragment has a fragment array relationship.

Then there is a component that depends on a property, which is an alias to the fragment array of an alias, and that second alias is an alias of the fragment. 

Easier explained with code:

```js
  model: null,

  firstAlias: Ember.computed.alias('model.someProp'),
  secondAlias: Ember.computed.alias('firstAlias.innerFragments'),

  someProps: function() {
    return this.get('secondAlias').mapBy('prop1');
  }.property('secondAlias.@each.prop1'),
```

There's an integration test in that demo app that invokes the bug.

jakesjews:
Cool thanks for the reproduction! I'll give it a look over when I get some free time

Amerr:
whats the status of the issue right now

[...]

PR #247, opened 2017-02-09
Fix for Max Call Stack Size Exceeded error when using multiple aliases

kobsy:
Possible fix for #190.

Borrowed from repo @DingoEatingFuzz wrote to illustrate the bug to write a failing test. The issue as far as we could tell seemed to result from infinite recursion from always generating false-positives when checking what keys changed. Since the model's `_attributes` assigned in `states.js` L173 was the fragment's `InternalModel`, the `isEqual` comparison to the Object returned from the server in Ember Data's `_changedKeys` method would always be false, and thus the key would endlessly be returned to be passed along to observers watching for a change ... which would cause them to call `setupData`, which would call `_changedKeys`, etc. The combination of assigning the updated data to `_attributes` and then filtering `_changedKeys` based on whether the returned attributes are actually equal (cheating, perhaps, in using `JSON.stringify`) seems to have resolved the problem.

### DIFF
--- a/addon/states.js
+++ b/addon/states.js
@@ -7,6 +7,7 @@ import RootState from 'ember-data/-private/system/model/states';
 
 const get = Ember.get;
 const create = Object.create || Ember.create;
+const assign = Ember.assign;
 
 const didSetProperty = RootState.loaded.saved.didSetProperty;
 const propertyWasReset = RootState.loaded.updated.uncommitted.propertyWasReset;
@@ -168,9 +169,9 @@ export default FragmentRootState;
 
 export function fragmentDidDirty(record, key, fragment) {
   if (!get(record, 'isDeleted')) {
-    // Add the fragment as a placeholder in the owner record's
+    // Add the fragment data as a placeholder in the owner record's
     // `_attributes` hash to indicate it is dirty
-    record._internalModel._attributes[key] = fragment;
+    record._internalModel._attributes[key] = fragment && assign({}, fragment._data, fragment._attributes);
 
     record.send('becomeDirty');
   }
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "ember-cli-dependency-checker": "^1.3.0",
     "ember-cli-eslint": "3.0.2",
     "ember-cli-htmlbars": "^1.1.1",
+    "ember-cli-htmlbars-inline-precompile": "^0.3.6",
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-pretender": "^1.0.1",
     "ember-cli-qunit": "^2.2.3",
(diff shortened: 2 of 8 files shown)


workmanw:
@kobsy Thank you!! I'll make time to review and test this tomorrow. We have another patch release coming for ember-data 2.12. If this get's merged, I'll be sure this gets in there.

addon/states.js
```diff
@@ -168,9 +169,9 @@ export default FragmentRootState;
 
 export function fragmentDidDirty(record, key, fragment) {
   if (!get(record, 'isDeleted')) {
-    // Add the fragment as a placeholder in the owner record's
+    // Add the fragment data as a placeholder in the owner record's
     // `_attributes` hash to indicate it is dirty
-    record._internalModel._attributes[key] = fragment;
+    record._internalModel._attributes[key] = fragment && assign(assign({}, fragment._data), fragment._attributes);
```
workmanw: Assign takes multiple sources, like so: `assign({}, fragment._data, fragment._attributes)`. Any reason to not do that?

kobsy: Oh! I hadn't realized that. No reason not to do it that way at all. 🙂

workmanw:
@kobsy Awesome. 👍 . 

--------

Hey @jakesjews can you give this a once over when you have some time. Maybe pull it into your app and make sure it doesn't bust any tests or anything ?

For reference, you can just use the following in your package.json. 
```json
 "ember-data-model-fragments": "lytics/ember-data-model-fragments#pull/247/head"
```

app/components/alias-component.js
```diff
@@ -0,0 +1 @@
+export { default } from 'ember-data-model-fragments/components/alias-component';
```
jakesjews: I think this export should be part of the test dummy app otherwise its going to be exported from the addon.

tests/dummy/app/components/alias-component.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  model: null,
+
+  firstAlias: Ember.computed.alias('model.passenger'),
+  secondAlias: Ember.computed.alias('firstAlias.name')
+
+  // someProps: function() {
```
jakesjews: can we remove these commented out lines?

package.json
```diff
@@ -33,6 +33,7 @@
     "ember-cli-dependency-checker": "^1.3.0",
     "ember-cli-eslint": "3.0.2",
     "ember-cli-htmlbars": "^1.1.1",
+    "ember-cli-htmlbars-inline-precompile": "0.3.6",
```
jakesjews: can we put a semver caret on this `^`

jakesjews:
@workmanw just some small nits but otherwise my application's tests still pass.

kobsy:
Cool. Good catches on those bits. I've amended the first commit to clean it up and add the semver caret.

jakesjews:
@kobsy thanks for the PR!

PR #252, opened 2017-03-01
Fixed `modelFactoryFor` bug. 

### LINKED ISSUE #251: Breaking change ED 2.11.1 to 2.11.2
Really useful package!

So I've got an app that was working fine with Model Fragments `2.11.1` and Ember Data `2.11.1`. However when I upgrade ED to `2.11.2` or `2.11.3` I get an error.

The error I'm hitting is this assertion:
https://github.com/lytics/ember-data-model-fragments/blob/master/addon/transforms/fragment.js#L59

With ED `2.11.1` the serializer for the fragment must be a `JSONSerializer` whereas with the upgraded ED it's a `JSONAPISerializer`.

My setup is that I've got a Fragment serializer:

```js
// app/pods/fragment/serializer.js
import Ember from 'ember';
import DS from 'ember-data';

const { String: { dasherize } } = Ember;
const { JSONSerializer } = DS;

export default JSONSerializer.extend({
  keyForAttribute(attr) {
    return dasherize(attr);
  }
});
```

And an initializer as per your README:

```js
// app/initializers/fragment-serializer.js
import FragmentSerializer from '../pods/fragment/serializer';

export function initialize(application) {
    application.register('serializer:-fragment', FragmentSerializer);
}

export default {
    name: 'fragment-serializer',
    initialize: initialize
};
```

Am I doing something wrong that shouldn't have worked in the first place? Or has the ED change introduced something breaking?

workmanw:
Fixes #251 .

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -368,12 +368,18 @@ function getFragmentTransform(owner, store, attributeType) {
   let _super = ContainerInstanceCachePrototype._fallbacksFor;
   ContainerInstanceCachePrototype._fallbacksFor = function _modelFragmentsPatchedFallbacksFor(namespace, preferredKey) {
     if (namespace === 'serializer') {
-      let model = this._store.modelFactoryFor(preferredKey);
-      if (model && Fragment.detect(model)) {
-        return [
-          '-fragment',
-          '-default'
-        ];
+      // ember-data-v2.11.2 changed `modelFactoryFor` to return a "factory manager".
+      // Test for that accordingly.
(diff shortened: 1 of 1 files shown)


jakesjews:
Nice! thats for fixing this up. LGTM

workmanw:
@jakesjews I'm going to go ahead and cut `2.11.2`. I don't foresee any more changes between now and 2.12 releasing. Sound good?

jakesjews:
Sounds good to me

PR #255, opened 2017-03-02
Upgrading to QUnit 2.0.

### LINKED ISSUE #253: Upgrade to QUnit 2.0 (ember-cli-qunit-3.x)
There are a number of reasons that we should do this. Primarily because QUnit 2.0 has the latest and greatest. Also because `ember-test-helpers` uses `lookupFactory` and that triggers a deprecation warning during the tests (not for actual runtime).

```javascript
DEPRECATION: Using "_lookupFactory" is deprecated. Please use container.factoryFor instead. [deprecation id: container-lookupFactory] See http://emberjs.com/deprecations/v2.x/#toc_migrating-from-_lookupfactory-to-factoryfor for more details.
  at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:16353:19)
  at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16464:7)
  at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:16383:12)
  at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16464:7)
  at invoke (http://localhost:4200/assets/vendor.js:16480:7)
  at deprecate (http://localhost:4200/assets/vendor.js:16448:32)
  at Object.deprecate (http://localhost:4200/assets/vendor.js:28289:37)
  at Container.lookupFactory (http://localhost:4200/assets/vendor.js:11882:19)
  at Klass.setupComponentIntegrationTest (http://localhost:4200/assets/test-support.js:7950:39)
```

So this deprecation comes from `ember-test-helpers` which is required by `ember-qunit` which is required by `ember-cli-qunit` :). To remove the warning, we must upgrade `ember-cli-qunit`. 

The one hang up we have is ember-dev doesn't support qunit 2.0. But I've written an addon to provide the same functionality `ember-qunit-assert-helpers`. We'll have to drop that in as well. 

This is probably going to be on the backburner for a little while. Just jotting down all my notes.

workmanw:
Fixes #253.

I decided to go ahead and do this because I wanted to validate the addon I built for this (`ember-qunit-assert-helpers`). It turned out to be incredibly easy.

The one thing to note is that `assert.expectDeprecation()` previously could be called at the beginning of a test, just as `assert.expect(10)`. However the implementation of `ember-qunit-assert-helpers` does not all this, due in part to QUnit changes. Obviously not a deal breaker, just something to note.


workmanw:
Ha! And as I say that, travis makes a liar out of me. Will investigate.

jakesjews:
@workmanw I wonder if it would be easier to make our tests use chrome or firefox instead of phantomjs. I've had lots of trouble in the past getting a recent phantom working on travis.

workmanw:
@jakesjews Yea when I did some quick googling I saw a few other people also running into to trouble with Phantomjs. But this particular failure seems like some kind of run time bug with the addon. It's so odd that I have to get to the bottom of it. Stay tuned :)

workmanw:
I have an update. The problem is actually caused by different versions of PhantomJS. Using the docker image locally I can reproduce this. I'm not yet sure why it works for 2.1.14 and not for 1.9.8, I'm going to investigate that because that is kind of concerning that maybe this addon has an unknown issue.

But it turns out that because the Travis CI image has 1.9.8 on the FS, when we kick off `ember try:one` it runs `npm install` which causes `node install.js` to "unbuild phantomjs-prebuilt@2.1.14” and use the default version available on the travis-ci container image (1.9.8).

It seems like adding `"phantomjs": "2.1.3"` to the `devDependencies` in my package.json resolves the issue (as it prevents npm from uninstalling it). So that's an easy workaround.

Another is to just use Chrome as @jakesjews mentioned. Ember-ajax for example is also doing the same ([.travis.yml](https://github.com/ember-cli/ember-ajax/blob/master/.travis.yml)). Jake, do you something similar to that .travis.yml to make Chrome work on Travis or is there another trick?

workmanw:
@jakesjew Turns out the problem was just that we needed to install phantomjs globally (-g) in our Travis.yml. This is what the ember-cli blueprint does for new apps/add-ons.

I still don't know why it failed on phantomjs 1.9, but I don't think it matters much given 2.1 is default.

package.json
```diff
@@ -36,15 +36,15 @@
...
+    "ember-cli-qunit": "^3.1.1",
     "ember-cli-release": "^1.0.0-beta.2",
     "ember-cli-shims": "^1.0.2",
     "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
     "ember-data": "~2.11.0",
-    "ember-dev": "emberjs/ember-dev",
     "ember-disable-prototype-extensions": "^1.1.0",
     "ember-load-initializers": "^0.6.0",
+    "ember-qunit-assert-helpers": "0.1.2",
```
jakesjews: do we want a semver caret on this?

workmanw:
@jakesjews Good catch on the semver caret.

jakesjews:
LGTM!

PR #265, opened 2017-05-09
revert fix-infinite-changed-key-loop

jakesjews:
Resolves https://github.com/lytics/ember-data-model-fragments/issues/259 but removes fix for https://github.com/lytics/ember-data-model-fragments/issues/190


workmanw:
:+1:

PR #270, opened 2017-06-11
WIP ember-data-model-fragments-2.14

workmanw:
Not quite ready to merge. Still waiting on: https://github.com/emberjs/data/pull/5008 to make it into beta. 

- [x] `fragment.ready` needs changed -- https://github.com/emberjs/data/commit/ff7893868560d3338c91c57ee0489d5ec120d542
- [x] Remove old backwards compat code.
- [x] Fix up ember-try scenarios. 

This is mostly ready. Need to wait for ember-2.14.0 to be released. Then I'll squash it all and submit for review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-data-model-fragments",
-  "version": "2.11.5",
+  "version": "2.14.0",
   "description": "Ember Data addon to support nested JSON documents",
   "keywords": [
     "ember-addon",
@@ -44,7 +44,7 @@
     "ember-cli-qunit": "^4.0.0",
     "ember-cli-shims": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
-    "ember-data": "~2.13.0",
+    "ember-data": "~2.14.1",
     "ember-disable-prototype-extensions": "^1.1.0",
     "ember-load-initializers": "^1.0.0",
     "ember-qunit-assert-helpers": "^0.1.3",
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ cache:
     - $HOME/.npm
 
 env:
-  - EMBER_TRY_SCENARIO=ember-2-11
-  - EMBER_TRY_SCENARIO=ember-2-12
+  # - EMBER_TRY_SCENARIO=ember-2-14
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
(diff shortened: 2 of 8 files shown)


workmanw:
Hey @jakesjews -- If you have time today, do you think you can give this a review. I just commented out the ember-2-14 scenario in the travis-ci file for now (because ember-2.14 hasn't shipped yet). The release scenario will test ember-2.13 w/ ember-data-2.14. 

I think at this point we should just release a new version of ember-data-model-fragments to keep from disrupting others. When ember-2.14 ships we can just uncomment the test scenario.

jakesjews:
@workmanw this looks like some nice cleanup so LGTM. My only concern is that maybe we should make this a semver major release in case anyone has a ^ in their package.json

package.json
```diff
@@ -44,7 +44,7 @@
     "ember-cli-qunit": "^4.0.0",
     "ember-cli-shims": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
-    "ember-data": "~2.13.0",
+    "ember-data": "2.14.1",
```
jakesjews: should we keep the ~ on here?

workmanw: Yea, we should definitely put the ~ in there. Good catch.

workmanw:
> My only concern is that maybe we should make this a semver major release in case anyone has a ^ in their package.json

Yea. I had thought of that too. I was ultimately torn because I like the fact that each of our version numbers correlate to ember-data versions. My hope was that if you were using `^` for ember-data-model-fragments, you'd also be using it for `ember-data`. Likewise with the `~`.

jakesjews:
@workmanw I think we have some gaps where the version number doesn't sync up with ember-data so maybe its not too bad to make a clean version break.

workmanw:
@jakesjews Sorry for the delay. Honestly it doesn't matter too much to me either. Do you know of any other addons that patch ember/ember-data like this? I was hoping to just find some precedence. Liquid-fire was the only one that came to mind and they do `0.X` releases (which is neither of the two proposals :P). 

I go back and forth between what would it look like if we were on v5 of model-fragments and there had been no new features in 2, 3, 4 or 5. Does that seem weird. Maybe I'm overthinking it.

jakesjews:
@workmanw no problem on the delay. It looks like https://github.com/danielspaniel/ember-data-change-tracker/ and https://github.com/mdehoog/ember-data-has-many-query do something sorta similar and they seem to follow semver.

jakesjews:
@workmanw after thinking about it I'm cool with the version as is. Do you wanna merge this in?

workmanw:
Hey @jakesjews -- Sorry. I got pulled into some work stuff for several days. I hope you don't think I was stalling because I didn't want to bump the semver.

workmanw:
Awesome 👍 I'm going to do a release now.

PR #274, opened 2017-08-09
Fix changedAttributes out of sync with hasDirtyAttributes

dwickern:
If I check `changedAttributes()` inside an observer for `hasDirtyAttributes`, the two values don't agree. The `changedAttributes` will say the property is dirty while `hasDirtyAttributes` says it's not, or vice-versa.


shidel-dev:
@dwickern I am experiencing this issue as well. Did you find a way to resolve this?

PR #283, opened 2017-12-01
Upgrade to ember 2.17

jakesjews:


jakesjews:
@workmanw I don't wanna bug you too much if your busy but does this PR look ok?

PR #286, opened 2018-01-06
When defaultValue is a function, don't copy

Techn1x:
Fixes https://github.com/lytics/ember-data-model-fragments/issues/281


jakesjews:
Thanks for the PR! Can you add a small test for this then It will be good to go.

Techn1x:
Sure! Although I just went to write a test for this, I got a bit stuck..

I am not quite sure how to test that a fragment's default value hasn't been copied when it's been defined as a function, because I can't really compare the difference, consider;
```javascript
let defaultValue = {
  first: 'Oath',
  last: 'Keeper'
};

let Sword = DS.Model.extend({
  name: MF.fragment('name', { defaultValue() { return defaultValue; } })
});

owner.register('model:sword', Sword);

let sword = store.createRecord('sword');
```

How can I assert that `sword`'s default value for `name` is or isn't a copy?

My webapp can tell the difference because I have something like the following, where `a` and `b` are objects that don't implement ember copyable, so the copy ends up broken/different
```javascript
let defaultValue = {
  a: moment(),
  b: moment.duration()
};
```

If someone has the time to help with writing this test that would be great :)

jakesjews:
It might work to add a test which does what you are describing with the objects that don't have ember.copyable and verifying that it doesn't throw an exception.

Techn1x:
@jakesjews sorry for taking so long to get to this! I've added tests that do exactly as you said - defaultValue function includes an object that doesn't implement Ember.Copyable, so that if it is deep copied an exception will be thrown / test fails.

Not sure why the build is failing the ember-default scenario in TravisCI, but the others scenario's passed. I don't think it's to do with this PR

jakesjews:
@Techn1x thanks a bunch! @workmanw it looks like the CI failures are related to some upcoming ember-data changes so I think its ok to merge this.

Issue #285, opened 2017-12-22
Tracking issue for removal of `ContainerInstanceCache`


rwjblue:
I just proposed a simplification for Ember Data, which removes the usage of `ContainerInstanceCache`.  This would be a breaking change to ember-data-model-fragments, so I wanted to open an issue proactively so we can track things...

See https://github.com/emberjs/data/pull/5300

rondale-sc:
Just ran into this.  Looks like at the moment ConstainterInstanceCache is used in the ED extensions: 

https://github.com/lytics/ember-data-model-fragments/blob/c7a57b3ee2d75287a395c49830e84b26d4ddfc70/addon/ext.js#L323

rondale-sc:
@rwjblue I believe this can be closed by:

https://github.com/lytics/ember-data-model-fragments/pull/287 

Can you confirm. :beers:

workmanw:
Bump. @rwjblue Can you take a quick peak and confirm this was resolved by @rondale-sc ? Should be a 5 second look.

workmanw:
Closing based on @rwjblue's comment in #287.

PR #287, opened 2018-01-26
Update to use Ember Data 3.0

rondale-sc:


jakesjews:
@rondale-sc looks awesome thanks for doing this! @workmanw does this look good to you?

workmanw:
👍 No concerns here.

rondale-sc:
@workmanw Do you have a trajectory for the next release to be cut from NPM.  I'd much prefer not to pin to a commit to get on 3.0 beta.  :beers:

workmanw:
Yea, let's make sure #285 is closable, then we'll cut a new release.

rwjblue:
Sorry for the delay, ya looks good to me.

workmanw:
@rondale-sc @jakesjews So we generally like to keep all of our ember packages all in lock step. However none of the 3 packages (ember, data, cli) have a v3 out yet. I think it would be a mistake for us to cut 3.0 right now while still using 2.18 packages. That said, we could release a 3.0.0-beta.1. I would be okay with that. Thoguths?

jakesjews:
I'm in agreement. I think a beta seems like a good idea.

rondale-sc:
@workmanw Totally.  I think pushing a beta now would be in perfect line with everything.  The Ember.js 3.0 beta moves to stable on Monday so this'll be perfect timing! :beers:

workmanw:
@rondale-sc 3.0.0-beta.1 was published to npm about an hour ago.

PR #297, opened 2018-05-15
Prevent infinite recursion on save

JosephHalter:
Prevent infinite recursion when saving a data containing a fragment which contains a fragment which contains a fragmentArray which has changed due to the way notifyPropertyChange works. setupFragment was calling setFragmentData which cause setupFragment to be called again, and never give it a change to assign fragment in internal model data.

Fixes a bunch of already reported but unsolved issues such as https://github.com/lytics/ember-data-model-fragments/issues/190 without any side effect. I did spend a lot of time to debug it, I'm sure a lot of people would like this to be merged.


JosephHalter:
Probably also fixes https://github.com/lytics/ember-data-model-fragments/pull/247

workmanw:
@JosephHalter Thanks a lot for this! As you noted, this has been a long running issue we've had trouble with. Would you mind creating a test for this infinite recursion scenario. It may be a little cumbersome, but it would really help us ensure it doesn't resurface.

Also, would you mind adding a code comment in the first if block (line ~100) that explains why the calls happen in a different order. Looking at the code it's unobvious to me. 

Thanks again!

JosephHalter:
I try adding a test on it but it pass even without my patch. However it definitely fails in my application with `Maximum call stack size exceeded.` using exactly the same use case, a model with a fragment containing a fragment containing a fragmentArray which has changed. When `_adapterDidCommit` is called it calls `setupFragment` which calls `setFragmentData` which does `content.replace(0, get(content, 'length'), processedData)` which triggers `notifyPropertyChange` which once again come into `setupFragment` and since we didn't assign the fragment in `_fragments` yet it calls once again `setFragmentData` instead of just returning, resulting in an infinite loop where the assignation is never done. I saw plenty of references to this bug so it's not just me: https://github.com/emberjs/ember.js/pull/13957 Using previous versions of ember-data or ember-data-model-fragments did nothing to solve it either.

I suspect the ember runloop just doesn't work the same way when in test. I don't know how to make a better test. The fix has however no side effect. After more than 1 day of debug, adding logs and breakthrough points everywhere to pinpoint the exact bug.

jakesjews:
Its ok with me to merge this with a comment explaining why the calls happen in different order

JosephHalter:
I added the comment and remove my attempt at adding a test on it since it didn't provide the expected value.

PR #300, opened 2018-07-13
Fix ember-data 3.x compability

JosephHalter:
Fixes https://github.com/lytics/ember-data-model-fragments/issues/290


jakesjews:
awesome thanks so much for this! @workmanw does this look good to you?

JosephHalter:
@jakesjews It was a pain to debug, had to do `ember build --environment=production` each time I changed some line, configure nginx in local and slowly find exactly when where it all went wrong and how to get the missing modelName using another way. I hope that it can be useful to someone else.

PR #301, opened 2018-07-14
fix import lint error

jakesjews:


jakesjews:
Good catch Ill put up a PR to fix

PR #302, opened 2018-07-16
replace deprecated exists-sync package with fs.existsSync

jakesjews:


jakesjews:
@workmanw does this look cool to you?

PR #303, opened 2018-07-16
Remove deprecated exports

jakesjews:


jakesjews:
@workmanw does this look ok to you? It was slated for removal in 3.0 so I figured we can drop it

PR #308, opened 2018-07-24
Use ember-copy instead of Ember.copy/Ember.Copyable

### LINKED ISSUE #307: Ember.copy / Copyable mixin deprecated
As of Ember 3.3, [Ember.copy and the Copyable mixin are deprecated](https://emberjs.com/blog/2018/07/16/ember-3-3-released.html#toc_deprecations-3).  This can lead to a large volume of deprecation warnings being dumped into the JS console.

Switching to the new [ember-copy addon](https://www.npmjs.com/package/ember-copy) addon clears up the deprecation warnings (see #308)

pauln:
Ember.copy/Ember.Copyable have been deprecated as of Ember 3.3, but the functionality has been made available via a new `ember-copy` addon.

Fixes #307.


jakesjews:
looks great @workmanw hows this look to you?

workmanw:
To be honest, I'm not sure about this. Do we call `.copy` internally or is that just a convenience for people who are using it at the application tier? 

My concern is if Ember is deprecating this capability (and planning to ultimately remove it), then we should too deprecate it. Suppressing these warnings makes it harder for users to find that they're using a deprecate feature.

jakesjews:
@workmanw looks like its used for getDefaultValue https://github.com/lytics/ember-data-model-fragments/blob/66ac3a6121111ca9b2409229251d161bdc365403/addon/attributes.js#L389

workmanw:
Oh it's right in the PR we use it. 🤦‍♂️ . I was more focused on the `Ember.Copyable` I think. Okay, this is fine with me.

package.json
```diff
@@ -49,6 +49,7 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-uglify": "^2.0.0",
     "ember-cli-update": "^0.14.0",
+    "ember-copy": "^1.0.0",
```
jakesjews: @pauln can you move this to regular dependencies and update the yarn.lock file

pauln: Done!

PR #309, opened 2018-07-25
Fix bug where setting array fragment attr to null can cause error

Techn1x:
Fixes https://github.com/lytics/ember-data-model-fragments/issues/306

EDIT: I'll add a test soon

### DIFF
--- a/addon/attributes.js
+++ b/addon/attributes.js
@@ -318,7 +318,7 @@ function fragmentArrayProperty(metaType, options, createArray) {
       assert('A fragment array property can only be assigned an array or null');
     }
 
-    if (internalModel._data[key] !== fragments || get(fragments, 'hasDirtyAttributes')) {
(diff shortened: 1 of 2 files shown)


jakesjews:
@Techn1x looks like some small lint errors for trailing spaces then I think this looks good

Techn1x:
No problem. The test I added is also slightly broken, will fix that up too (unfortunately for some reason `ember test` is broken on my dev machine for this addon..)

jakesjews:
@workmanw this looks good to me how about you?

workmanw:
@Techn1x :+1: for the great test and fix. I think Ember's copy is deprecated right? And we just started using `ember-copy` instead. Can you rebase on to master and check that. If so, I think ti's as easy as swapping the import.

Techn1x:
Thanks guys. Have done the rebase and now using ember-copy

PR #310, opened 2018-07-31
Fix EDMF for use with >3.2

rondale-sc:
The `getRecords` function inside Ember Data treats attributes
differently.

Where once the it would set each attribute on the internal model with
`options[]=` (triggering the setter/getter we install to properly
install EDMF) it now calls `InternalModel#setDirtyAttribute`.

The deciding factor for whether something is an `attribute` was based
off the `meta` set on the installed CP's.  We're instead telling the
`InternalModel#fields` function to properly recognize fragments as
`"fragments"` which bypasses the `setDirtyAttribute` call and instead
does the default behavior from before.

Reopening the Model in this way preserves the previous behavior.

I suspect, this will likely need a large refactor when RFC293 lands

https://github.com/igorT/rfcs/blob/17e8f97049452ae18d3b12ee23c4858a6d5e0bfd/text/0000-model-data.md

Luckily, I believe IgorT is already begun the work there with this spike
listed in the above RFC:

https://github.com/igorT/ember-data.model-fragments/tree/igor/model-data

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -10,6 +10,7 @@ import {
 } from './fragment';
 import FragmentArray from './array/fragment';
 import { isPresent } from '@ember/utils';
+import { computed } from '@ember/object';
 
 function serializerForFragment(owner, normalizedModelName) {
   let serializer = owner.lookup(`serializer:${normalizedModelName}`);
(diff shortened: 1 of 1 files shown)


rondale-sc:
The specific part we are overriding to do the default behavior is this switch statement:

https://github.com/emberjs/data/blob/d41601529a09636e636524a6d80b4d877952c344/addon/-legacy-private/system/model/internal-model.js#L381

jakesjews:
@rondale-sc awesome it's so nice seeing a green build! @workmanw hows this look to you?

rwjblue:
Seems reasonable to me (FWIW)...

workmanw:
@rondale-sc I assume the model-fragment tests were broken with ember-data 3.2 prior to this fix? Just want to make sure we have some kind of test coverage.

rwjblue:
Yes, a number of tests failed before this PR when ran against ember-data@3.2+.

rondale-sc:
@workmanw @rwjblue Yeah, what Robert said is correct.  Specifically we had some tests around fragments being handled when a POJO with a fragment is passed to createRecord (which was what started me on this path).

PR #319, opened 2018-11-05
Return an array with the diff in changedAttributes

### LINKED ISSUE #318: snapshot.changedAttributes() is broken with fragments and Ember Data > 3.1
In Ember Data they have removed [the usage of Ember.copy](https://github.com/emberjs/data/commit/23a87e8e7a7cd39e33ef877148f7d9e8f103c2ab#diff-67322a0874ff9f30dbdfc17e18916a4dR174) so when you use `snapshot.changedAtributes()` it expects that `this._changedAttributes[key]` is always an Array (now they are copying an element with slice instead of using Ember.copy).

The problem is with [this code of EDMF](https://github.com/lytics/ember-data-model-fragments/blob/master/addon/ext.js#L164-L170): it returns a Boolean instead of an Array. What's the utility of this change? 
I'm going to remove the `changedAttribute` method from this `model.reopen` to see what happens.

Gorzas:
Fixes #318

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -165,7 +165,10 @@ Model.reopen({
       // An actual diff of the fragment or fragment array is outside the scope
       // of this method, so just indicate that there is a change instead
       if (name in internalModel._attributes) {
-        diffData[name] = true;
+        diffData[name] = [
+          diffData[name][0],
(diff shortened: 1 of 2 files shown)


jakesjews:
@Gorzas thanks a bunch! @workmanw this looks good to me are you ok with merging?

PR #320, opened 2018-11-07
Nested polymorphic fragments do not respect custom Serializers normalize

DocX:
**Problem:**
In our applications we use polymorphic fragment models. Our application structure is complex, and hence our model types looks like `common/blocks/text`, which is specific to our ember application and different from the API. We want to keep the server API clean and persist types like `text`.

We found out that we can use custom ember `Serializer` with overriden `normalize` method, that can add inferred ember model type from the type in the payload. For example:

```
normalize(modelClass, resourceHash) {
    let result = this._super(modelClass, resourceHash);
    
    Object.assign(result.data.attributes, resourceHash, {
      $type: `common/blocks/${resourceHash.type}`
    });

    return result;
  }
```

This works perfectly when we have only flat structure of polymorphic models.

However it brakes when we expand the structure to be nested. For example we want to have polymorphic array of `common/content`, and then one of them, for example `common/content/with-block` includes another polymorphic array of `common/block`:

```js
// common/content/with-block model
blocks: MF.fragmentArray('common/block', { polymorphic: true, typeKey: '$type' })

// content-container model
contents: MF.fragmentArray('common/content', { polymorphic: true, typeKey: '$type' })
```

In this branch I've simulated the use case in the test cases, so we can discuss whether or not this is supposed to be supported by this package and how :)

### DIFF
--- a/tests/integration/nested_polymorphic_test.js
+++ b/tests/integration/nested_polymorphic_test.js
@@ -1,0 +1,177 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import Animal from 'dummy/models/animal';
+import Lion from 'dummy/models/lion';
+import PreyGiraffe from 'dummy/models/prey-giraffe';
+import PreyHog from 'dummy/models/prey-hog';
+import Pretender from 'pretender';
+
+let store, owner, server;
+
+module('integration - Nested Polymorphic fragments with custom type normalizer', function(hooks) {
+  setupApplicationTest(hooks);
+
+  hooks.beforeEach(function(assert) {
+    owner = this.owner;
+    store = owner.lookup('service:store');
+    server = new Pretender();
+
+    assert.expectNoDeprecation();
+  });
+
+  hooks.afterEach(function() {
+    store = null;
+    owner = null;
+    server.shutdown();
+  });
+
+  test('when all levels types are normalized', function(assert) {
+    let data = {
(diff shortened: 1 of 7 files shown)


tests/integration/nested_polymorphic_test.js
```diff
@@ -0,0 +1,177 @@
...
+      animals: [
+        {
+          _normalizeBackendTypeTo$Type: true,
+          backendType: 'lion',
+          name: 'Mittens',
+          hasManes: 'true',
+          preys: [
+            {
+              $type: 'prey-giraffe',
+              neckLength: 20
```
DocX: This case works correctly, as the normalization is not needed to be nested. It seems like the normalization is only respected on the root model level (zoo), and not correctly recursively applied to the nested fragments.

GavinJoyce:
A question for the maintainers, do you think that this is a bug that you would accept a PR for? Our alternative solution is to change we serialise our payloads on the server to include correct types, but we'd be happy to try and fix this in `ember-data-model-fragments` if you think that it's a bug

jakesjews:
I think a PR will be good for this. Would it include adding the typeKey meta fields to fragment properties? If so could it use something like [keyForPolymorphicType](https://www.emberjs.com/api/ember-data/3.5/classes/DS.RESTSerializer/methods/keyForPolymorphicType?anchor=keyForPolymorphicType) in the serializer?

PR #329, opened 2019-03-21
support an optional fragmentUniqueKey from the serializer

shidel-dev:


shidel-dev:
see #329 for details on the use case for this pull request

shidel-dev:
I am curious if this would be better on the model definition, or is the serializer the best place to configure this, as it is responsible for reconciling responses?

jakesjews:
The serializer seems like the best place. Could you add in a quick test for this functionality?

PR #352, opened 2019-12-22
Repairing tests

basz:
Hi, 

following up on https://github.com/lytics/ember-data-model-fragments/issues/338 I decided we should first get the tests working again. To do that I've reviewed all dependencies and most importantly replaced the deprecated ember-cli-qunit with ember-qunit, which solved a deprecation warnings while running the tests.

Final note, support for node 6 was removed in ember 14, so I took it upon myself to remove that one too from the 'engines' property in packages.

Edit: ember-lts-2.16 and ember-lts-2.18 aren't supported in v4 according to the list. So I removed them too from the test runs


basz:
obviously, anything above 3.12 fails at the moment

jakesjews:
LGTM @workmanw what do you think?

PR #380, opened 2020-10-05
Feature/379 update blueprint

Gorzas:


package.json
```diff
@@ -68,5 +68,8 @@
   },
   "ember-addon": {
     "configPath": "tests/dummy/config"
+  },
```
patocallaghan, line 71: I don't think we should add the edition in the package.json until we update the version of ember-cli (we're currently on ember-cli 3.3)

patocallaghan: FYI I've started trying to upgrade ember-cli in https://github.com/lytics/ember-data-model-fragments/pull/385

blueprints/fragment/files/app/__path__/__name__.js
```diff
@@ -1,6 +1,6 @@
-import DS from 'ember-data';
-import MF from 'ember-data-model-fragments';
+<%= importedModules.length ? `import { ${importedModules} } from '@ember-data/model';` : '' %>
+import Fragment from 'ember-data-model-fragments/fragment';
 
-export default MF.Fragment.extend({
+export default class <%= classifiedModuleName %> extends Fragment {
   <%= attrs %>
```
patocallaghan, line 5: Could we remove the space here because having it indented results in the first `attr` being indented too far?

```suggestion
<%= attrs %>
```

package.json
patocallaghan: @Gorzas if you pull the latest changes from master you should be able to remove these changes now

patocallaghan (changes requested):
Just some final touch ups and we should be good to go 💪

blueprints/fragment/files/app/__path__/__name__.js
Gorzas: @patocallaghan doing some tests with the generator, I've found that the result with attributes isn't valid. Look what happens when I type `ember g fragment foo bar:string`:

```
import { attr } from '@ember-data/model';
import Fragment from 'ember-data-model-fragments/fragment';

export default class Foo extends Fragment {
  bar: attr('string')
}
```

So I'm going to fix that too.

PR #381, opened 2020-10-09
fix(hasDirtyAttributes): fix hasDirtyAttributes when resetting a property

VincentMolinie:
# Steps to reproduce the issue
- Changing a fragment
- Then changing a property
- And finally resetting the property to its old value

# Expected behaviour
`hasDirtyAttributes` should be `true`

# Current behaviour
`hasDirtyAttributes` is `false`

PR #383, opened 2020-10-26
show test failing with hasChangedAttributes

GuillaumeCisco:


patocallaghan:
@GuillaumeCisco Apologies I was on vacation for the last week. I wonder if you could put your potential dummy fix in this PR too and maybe update the description? If anything to generate discussion about how to properly fix it. I don't know a whole lot about the internals of RecordData but we can ping the relevant folks to get them to look

tests/unit/fragment_array_property_test.js
```diff
@@ -140,20 +140,25 @@ module('unit - `MF.fragmentArray` property', function(hooks) {
         region: 'Crownlands',
         country: 'Westeros'
       });
-      assert.deepEqual(person.changedAttributes().addresses, undefined, 'a frehsly pushed object has no changes');
+      assert.deepEqual(person.changedAttributes().addresses, undefined, 'a freshly pushed object has no changes');
+      assert.ok(!person._internalModel.hasChangedAttributes(), 'a freshly pushed object has no changed attributes');
```
patocallaghan, line 144: Does `hasDirtyAttributes` give us the same results as `person._internalModel.hasChangedAttributes()`?

GuillaumeCisco:
> @GuillaumeCisco Apologies I was on vacation for the last week. I wonder if you could put your potential dummy fix in this PR too and maybe update the description? If anything to generate discussion about how to properly fix it. I don't know a whole lot about the internals of RecordData but we can ping the relevant folks to get them to look

The PR with the fix can be found on my fork here: https://github.com/GuillaumeCisco/ember-data-model-fragments/pull/1
In fact, it looks like to me `hasChangedAttributes` has never been correctly handled... It looks like a copy/paste from the original ember-data.
I don't know neither about how works RecordData so I cannot provide a smarter fix, help is welcome here :)

tests/unit/fragment_array_property_test.js
GuillaumeCisco: I don't even know. I think it should by reading the [documentation](https://guides.emberjs.com/v3.22.0/models/creating-updating-and-deleting-records/)
But `hasChangedAttributes` is an internal method, and this is the one we need to update for fixing the bug in my case.
As `hasChangedAttributes` will be called as I developed it in earlier post.

patocallaghan:
@GuillaumeCisco I think we can close this PR. It appears that the fix in https://github.com/lytics/ember-data-model-fragments/pull/381 makes the test you created here pass. They also include a fix with their test.

patocallaghan:
The fix from #381 has been released in [v5.0.0-beta.1](https://github.com/adopted-ember-addons/ember-data-model-fragments/releases/tag/v5.0.0-beta.1)

Issue #375, opened 2020-08-19
5.0.0-beta.0 - normalization incomplete


basz:
I have a model with a cloudDrive status fragment that in turn has two more fragments. called context and failure.

Over a webstomp connection I receive updates to my models via `store.pushPayload(payload)`. The model correctly updates except if it does not exists in the store. When it does exist in the store all properties are correctly updated except for the nested fragments... (note the state property is correctly updated.

"status" : {
  "state": "checked-out",
  "context": {
    "adapter": "dropbox",
    "identity_id": "xxx-xxx-xxx-xxx-xxx",
    "order_number": "2.4",
    "order_status": "cad",
    "checkout_time": "2020-08-19T11:51:26.000Z"
  },
  "failure": {
    "message_code": null,
    "message": null,
    "details": []
  }
}

So. a button instructs the backend to perform a 'check-in'... This might fail due to some reason and the whole model is transmitted via webstomp as a json API payload. Pushing that payload does not update the failure fragment.

Any ideas where to look/debug. I have confirmed the received payload is correct.

```
"data": {
    "type": "dossier/order",
    "id": "xxx-xxx-xxx-xxx-xxx",
    "attributes": {
      // redacted
      "cloud-drive": {
        "state": "checked-out",
        "context": {
          "checkout_time": "2020-08-19T11:51:26+00:00",
          "identity_id": "xxx-xxx-xxx-xxx-xxx",
          "adapter": "dropbox",
          "order_status": "cad",
          "order_number": "2.4"
        },
        "failure": {
          "message_code": "dossier.checkInFailure.noMutationsDetected",
          "message": "No changes where detected. Has the cloud drive been fully synchronized?",
          "details": []
        }
      }
    },
  }
}
```

basz:
setting a fragment to null on a model seems to be the culprit here. I would guess the fragment is then either deleted or recreated from the defaultValue. Is that indeed what I may expect to happen?

patocallaghan:
@basz I think I'm seeing a similar issue to you. While it's not exactly the same it has the same characteristic where the fragment isn't populated after being `null` initially.

In our case we are creating a record which is only sparsely populated but the rest of the attributes are filled in by the response from the backend.

I've created a failing test in https://github.com/lytics/ember-data-model-fragments/pull/384 to replicate it and have confirmed it did work before the 5.0 refactor (my tests passed).

patocallaghan:
[...]

PR #384, opened 2020-11-05
Make sure we can update fragments/fragment arrays after they are initially set to null

patocallaghan:
This is a failing test which mimics the bug in https://github.com/lytics/ember-data-model-fragments/issues/375. This worked before the 5.0.0 beta.

I originally commented most of this in https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/375#issuecomment-940076666. Essentially if you have a model/fragment which has a fragment array/fragment property set to `null` it will never update when a http response has those properties updated. I have confirmed running these failing tests on a v4 branch pass just fine.

I've tracked the bug down to the following lines of code in the `FragmentRecordData#didCommit` hook:

https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/4f77b2d5ea08685230ec60b5a3ddd2b3d3a7ec6c/addon/record-data.js#L426-L433

Basically because the existing property is set to `null` the `fragment = this.serverFragments[key];` evaluates to `null`. This means the `if (fragment)` check fails and we don't commit the new data to the fragment with `fragment._didCommit(attributes[key]);`.

So I've spiked out a potential solution but I'm unsure if this should be the direction I should be taking. I don't have as much context how Model Fragments/RecordData works so any pointers would be appreciated. Essentially we need to recreate the fragment or fragmentArray if it is `null` on the record and then re-set it on the record. The API of `createFragment`, for example, is `createFragment(store, declaredModelName, record, key, options, data)` but I can't find any possible way to get access to the `record` at this point in the stacktrace so I need to do some hacky stuff to get access to the host record, i.e. setting the `this._record` class property which, to be honest, seems a bit janky.

image

It's also worth pointing out that I've ran this particular change against the Intercom test suite which has thousands of model fragment related tests and it works.

I've peppered comments throughout the source but this is more for comprehension for the reader. If I get an approval I'll remove any comments I don't think are necessary.

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -27,18 +27,21 @@ export default class FragmentRecordData extends RecordData {
       super(identifier.type, identifier.id, identifier.clientId, store);
     }
 
+    // @patocallaghan - We need to keep a copy of the record so we can recreate the fragment/fragmentArray.
+    this._record = null;
     this.fragmentData = Object.create(null);
     this.serverFragments = Object.create(null);
     this.inFlightFragments = Object.create(null);
 
     this.fragments = Object.create(null);
-    this.fragmentNames = [];
+    // @patocallaghan - We need to keep the fragment definitions for later on in case we need recreate the fragment or fragmentArray
+    this.fragmentDefs = Object.create(null);
     let defs = store.attributesDefinitionFor(identifier.type, identifier.id, identifier.lid);
 
     Object.keys(defs).forEach(key => {
       let options = defs[key];
       if (options.isFragment) {
-        this.fragmentNames.push(key);
+        this.fragmentDefs[key] = defs[key];
       }
     });
     fragmentRecordDatas.set(identifier, this);
@@ -54,6 +57,8 @@ export default class FragmentRecordData extends RecordData {
   }
 
   setupFragment(key, options, declaredModelName, record) {
+    // @patocallaghan -  This is extremely janky way of making sure we always have a copy of the record saved. There must be a better way of doing this?
+    this._record = record;
     let data = this.getFragmentDataWithDefault(key, options, 'object');
     let fragment = this.fragments[key];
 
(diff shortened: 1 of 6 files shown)


addon/record-data.js
```diff
@@ -27,18 +27,21 @@ export default class FragmentRecordData extends RecordData {
...
 
+    // @patocallaghan - We need to keep a copy of the record so we can recreate the fragment/fragmentArray. There must be a better way to do this?
+    this._record = null;
     this.fragmentData = Object.create(null);
     this.serverFragments = Object.create(null);
     this.inFlightFragments = Object.create(null);
 
     this.fragments = Object.create(null);
-    this.fragmentNames = [];
+    // @patocallaghan - We need to keep the fragment definitions for later on in case we need recreate the fragment or fragmentArray
```
patocallaghan, line 37: `fragmentDefs` are used in https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/384/files#diff-a559085745304ba03c26083bacebf245f8ef7540dc430c48fcdb95327c49e5e1R445 when recreating the fragments/fragment arrays. There's no point keeping around `fragmentNames` as they can be inferred from `fragmentDefs`.

addon/record-data.js
```diff
@@ -27,18 +27,21 @@ export default class FragmentRecordData extends RecordData {
       super(identifier.type, identifier.id, identifier.clientId, store);
     }
 
+    // @patocallaghan - We need to keep a copy of the record so we can recreate the fragment/fragmentArray.
+    this._record = null;
```
patocallaghan, line 31: I couldn't find some "happy path" to retrieve a record from `RecordData` so I resort to setting this private field throughout the various hooks where the record is passed through.

patocallaghan:
@knownasilya @ro0gr any thoughts on this? It's another `null` related bug where if a fragment/fragment array is `null` and a non-null server response comes in, it never gets updated.

patocallaghan:

PR #390, opened 2020-12-10
Fix store push with `null` value

bobisjan:


addon/record-data.js
```diff
@@ -231,15 +231,19 @@ export default class FragmentRecordData extends RecordData {
           this.fragmentData[name] = data.attributes[name];
           let serverFragment = this.serverFragments[name];
           if (serverFragment) {
-            let fragmentKeys = [];
-            if (serverFragment instanceof StatefulArray) {
-              serverFragment.setupData(data.attributes[name]);
+            if (data.attributes[name] === null) {
+              delete this.serverFragments[name];
```
bobisjan, line 235: TBH I'm not familiar with the internals, hence I'm not sure if it is a proper fix 😞. At least I've run this test against v4.0 and the output is now same as before.

dwickern:
The test from this PR passes on the master branch.

I believe it's covered now by these two tests:
https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/d563e771c21c5d2331f67c585799ed36a2de977b/tests/unit/fragment_test.js#L429-L456
https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/d563e771c21c5d2331f67c585799ed36a2de977b/tests/unit/fragment_array_test.js#L370-L417

PR #397, opened 2021-04-20
fix: use send('loadedData') instead of loadedData method

### LINKED ISSUE #396: ModelFragments broken against EmberData canary
We recently did a cleanup pass to remove unneeded methods from InternalModel, it looks like model fragments may have been using a few of these. The refactor should be simple and backwards compatible. You can see each removed method or property and the appropriate change to make here: https://github.com/emberjs/data/pull/7453

Side note: we didn't notice this regression until now because the model-fragments partner nest was still set to allowed-fail. I noticed this on a different PR after re-activating the job. Unfortunately these PRs (the re-activation and the removal of unneeded methods) landed at the same time and so didn't catch this for a few days.

runspired:
resolves #396

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -236,7 +236,7 @@ Model.reopen({
 });
 
 Model.reopenClass({
-  fields: computed(function () {
+  fields: computed(function() {
     let map = new Map();
 
     this.eachComputedProperty((name, meta) => {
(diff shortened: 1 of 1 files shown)


patocallaghan:
Thanks! Released in https://github.com/adopted-ember-addons/ember-data-model-fragments/releases/tag/v5.0.0-beta.2

PR #404, opened 2021-08-02
Fix fragment attributes in save response being ignored

dwickern:
Steps to reproduce:
1. Modify a fragment attribute and save
2. The server returns a different value for the attribute in its response

**Expected:** the fragment attribute should have the new value from the server (consistent with regular ED attributes outside of fragments)
**Actual:** the fragment attribute still has the locally modified value

The root cause seems to be `willCommit` not being called for fragments. `willCommit` is normally responsible for moving `_attributes` into `_inFlightAttributes`:
https://github.com/emberjs/data/blob/f56b22c4d6eefab8915ca48979f7d813f612971d/packages/record-data/addon/-private/record-data.ts#L120-L123

This is important because `_changedKeys` prefers to take `_attributes` over attributes in the response body:
https://github.com/emberjs/data/blob/f56b22c4d6eefab8915ca48979f7d813f612971d/packages/record-data/addon/-private/record-data.ts#L763-L769

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -85,6 +85,11 @@ assign(RecordDataPrototype, {
   },
 
   didCommit(data) {
+    if (this._attributes) {
+      // willCommit was never called
+      this._inFlightAttributes = this._attributes;
+      this._attributes = null;
+    }
     this._isNew = false;
     if (data) {
       if (data.relationships) {
@@ -105,8 +110,7 @@ assign(RecordDataPrototype, {
 
     const changedKeys = this._changedKeys(data);
 
-    assign(this._data, this.__inFlightAttributes, this._attributes, data);
-    this._attributes = null;
+    assign(this._data, this._inFlightAttributes, data);
     this._inFlightAttributes = null;
     this._updateChangedAttributes();
 
--- a/tests/integration/save_test.js
+++ b/tests/integration/save_test.js
@@ -113,6 +113,53 @@ module('integration - Persistence', function(hooks) {
     });
   });
 
+  test('overwrite current state with fragment attributes from the save response', function(assert) {
+    return run(() => {
+      store.push({
+        data: {
+          type: 'person',
+          id: 1,
+          attributes: {
+            title: 'Lord',
+            name: {
+              first: 'Tyrion',
+              last: 'Lannister'
+            }
+          }
+        }
+      });
+
+      server.put('/people/1', (request) => {
+        const body = JSON.parse(request.requestBody);
+        assert.equal(body.person.title, 'modified');
+        assert.equal(body.person.name.first, 'modified');
+        assert.equal(body.person.name.last, 'modified');
+        body.person.title = 'Ser';
+        body.person.name.first = 'Tywin';
+        body.person.name.last = 'Lannister';
+        return [200, { 'Content-Type': 'application/json' }, JSON.stringify(body)];
+      });
+
+      return store
+        .find('person', 1)
+        .then(person => {
+          person.set('title', 'modified');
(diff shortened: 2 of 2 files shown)


patocallaghan:
@dwickern I came across this same issue myself in #418. Have you been using this fix in a fork or anything?

addon/ext.js
```diff
@@ -85,6 +85,11 @@ assign(RecordDataPrototype, {
   },
 
   didCommit(data) {
+    if (this._attributes) {
```
patocallaghan, line 88: I'm going to test this fix out in our app but should we be somehow making `willCommit` fire as that is the root cause?

knownasilya:
@patocallaghan if you can test this out, I'll merge it if it's good.

addon/ext.js
patocallaghan: So I had a look into whether we could make `willCommit` fire as expected and it's a bit all over the shop.

The [`willCommit`](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/1403665eb8b8a3e43c31a8adccc1d301bed26d8f/addon/record-data.js#L281) hook here works by iterating over any fragments registered on the record by looking at the `this.fragments` class field. Unfortunately it appears that we only add fragments to `this.fragments` in cases where we specifically [`set` a fragment/fragment array](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/1403665eb8b8a3e43c31a8adccc1d301bed26d8f/addon/record-data.js#L164), but not when we initialise them.

I did a small spike where I also push fragments/fragment arrays into `this.fragments` in [`setupFragment`](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/1403665eb8b8a3e43c31a8adccc1d301bed26d8f/addon/record-data.js#L56) and  [`setupFragmentArray`](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/1403665eb8b8a3e43c31a8adccc1d301bed26d8f/addon/record-data.js#L187), which are what are used when bootstrapping fragments. This works correctly and the `willCommit` hooks fire as expected and the fragments get their values updated from the server responses.

A problem I'm now seeing as a result of this change is that  `this.fragments` is used to calculate `changedAttributes`. My spike then causes all fragments to appear in `changedAttributes()` even just after a model with fragments has just been initialized. So I think a deeper refactor here would need to be figured out if we wanted `willCommit` to be called properly.

@knownasilya I'm going to test this specific fix in my app now to see whether it fixes our problem. If it does it would be a good fix to unblock an upgrade for us

dwickern:
@patocallaghan I've patched it locally:
```js
// app/initializers/ember-data-model-fragments-workaround.js
import { RecordData } from 'ember-data/-private';

export function initialize() {
  const didCommit = RecordData.prototype.didCommit;
  RecordData.prototype.didCommit = function didCommitWorkaround(data) {
    if (this._attributes) {
      // willCommit was never called
      this._inFlightAttributes = this._attributes;
      this._attributes = null;
    }
    return didCommit.call(this, data);
  };
}

export default {
  initialize,
  after: 'fragmentTransform', // after ember-data-model-fragments is initialized
};
```

patocallaghan:
Thanks @dwickern 👍

@knownasilya I've verified this fixes the issue in our app so I think this fix will do for now to unblock some folks 👍

addon/ext.js
dwickern: I spiked out a deeper refactor in #417

patocallaghan: Nice work @dwickern. I'll take a closer look at that PR when I get a chance at some point this week

patocallaghan:

PR #408, opened 2021-09-13
Update README syntax

charlesfries:


README.md
```diff
@@ -223,25 +222,22 @@ Ember Data attributes [support a `defaultValue` config option](http://emberjs.co
...
+  @array('string') titles;
+}
 ```
 
 Since JavaScript objects and arrays are passed by reference, the value of `defaultValue` is copied using `Ember.copy` in order to prevent all instances sharing the same value. If a `defaultValue` option is not specified, `fragment` properties default to `null` and `fragmentArray` properties default to an empty array. Note that this may cause confusion when creating a record with a `fragmentArray` property:
 
 ```javascript
 let person = store.createRecord('person');
-let addresses = person.get('addresses'); // null
+let addresses = person.addresses; // null
```
knownasilya, line 240: I think `get` is basically still required on models, since attrs can be relationships.

README.md
```diff
@@ -253,46 +249,49 @@ Like `attr`, the `defaultValue` option can be a function that is invoked to gene
...
-export default Model.extend({
-  name: fragment('name', {
+export default class PersonModel extends Model {
+  @fragment('name', {
     defaultValue() {
       return {
         first: 'Unsullied',
-        last: Ember.uuid()
-      }
+        last: Ember.uuid(),
```
knownasilya, line 261: Should probably not use a global Ember in an example. Can import `uuid` and use that (make a comment about installing uuid from npm)

charlesfries: Added the `uuid` note, though it still seems kind of awkward. Do you know if it's possible to use `guidFor` here?

`import { guidFor } from '@ember/object/internals';`

knownasilya: Maybe just change the example 🙂

knownasilya (approved):

knownasilya:
@charlesfries looks like there are some conflicts again. I'll merge as soon as those are resolved.

PR #415, opened 2021-11-30
[breaking][bugfix] respect null on server updates 

ro0gr:
Prior to this change, if a `fragment` was initialized with some value, and then server responds with a `null` it did not have any impact on a fragment state, and it kept having a non-empty value assigned. Now it's fixed, so the fragment attribute can be updated to `null`.

The same behavior is applied to the `fragmentArray`. Technically, this is a *breaking change*, since before, if non empty
`fragmentArray` received a `null` from the server, instead of nullifying the attribute, the `fragmentArray` instance just
became empty(with no items inside).

It was misaligned with how the `fragmentArray` handles `null` on create, cause during creation it is nullable. So for consistency, we let it be nullable on updates as well.

### DIFF
--- a/tests/unit/fragment_array_test.js
+++ b/tests/unit/fragment_array_test.js
@@ -282,4 +282,66 @@ module('unit - `MF.fragmentArray`', function(hooks) {
       });
     });
   });
+
+  test('can be created with null', function(assert) {
+    let person = store.push({
+      data: {
+        type: 'person',
+        id: 1,
+        attributes: {
+          names: null
+        }
+      }
+    });
+
+    run(() => {
+      assert.strictEqual(person.names, null);
+    });
+  });
+
+  test('can be updated to null', function(assert) {
+    let person = store.push({
+      data: {
+        type: 'person',
+        id: 1,
+        attributes: {
+          names: [
+            {
+              first: 'Catelyn',
+              last: 'Tully'
+            },
+            {
+              first: 'Catelyn',
+              last: 'Stark'
+            }
+          ]
+        }
+      }
+    });
+
+    run(() => {
+      assert.deepEqual(person.names.toArray().map((f) => f.serialize()), [
+        {
+          'first': 'Catelyn',
+          'last': 'Tully'
+        },
+        {
+          'first': 'Catelyn',
+          'last': 'Stark'
+        }
+      ]);
+
+      store.push({
+        data: {
+          type: 'person',
+          id: 1,
+          attributes: {
+            names: null
+          }
+        }
+      });
+
+      assert.strictEqual(person.names, null);
+    });
+  });
 });
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -232,7 +232,11 @@ export default class FragmentRecordData extends RecordData {
           let serverFragment = this.serverFragments[name];
           if (serverFragment) {
             let fragmentKeys = [];
-            if (serverFragment instanceof StatefulArray) {
+            if (data.attributes[name] === null) {
+              // if we have data with a Fragment set up, but now we've received null,
+              // delete the null fragment from serverFragments.
+              delete this.serverFragments[name];
+            } else if (serverFragment instanceof StatefulArray) {
               serverFragment.setupData(data.attributes[name]);
             } else {
               fragmentKeys = serverFragment._internalModel._recordData.pushData({ attributes: data.attributes[name] }, calculateChange);
--- a/tests/unit/fragment_test.js
+++ b/tests/unit/fragment_test.js
@@ -307,4 +307,51 @@ module('unit - `MF.Fragment`', function(hooks) {
       assert.ok(person.get('names').isEvery('readyWasCalled'), 'when creating model that has fragmentArray');
     });
   });
+
+  test('can be created with null', async function(assert) {
+    let person = store.push({
+      data: {
+        type: 'person',
+        id: 1,
+        attributes: {
+          name: null
+        }
+      }
+    });
+
+    return run(() => {
+      assert.strictEqual(person.name, null);
+    });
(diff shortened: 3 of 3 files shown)


tests/integration/save_test.js
```diff
@@ -465,6 +465,39 @@ module('integration - Persistence', function(hooks) {
...
+        data: {
+          type: 'person',
+          id: 1,
+          attributes: {
+            name: null
+          }
+        }
+      });
+
+      assert.equal(person.name, null);
```
patocallaghan, line 497: @ro0gr This looks like a good change. Any chance you could update the test to test `person.names` too so that we know this works for fragment arrays too?

ro0gr: sure, will do!

tests/unit/fragment_array_test.js
```diff
@@ -282,4 +282,66 @@ module('unit - `MF.fragmentArray`', function(hooks) {
...
+        data: {
+          type: 'person',
+          id: 1,
+          attributes: {
+            names: null
+          }
+        }
+      });
+
+      assert.strictEqual(person.names, null);
```
ro0gr, line 344: so the current FragmentArray behavior in the master is that `person.names` are equal to an empty fragment array wrapper:
```
      assert.deepEqual(person.names.toArray().map((f) => f.serialize()), []);
```
So the change from this PR is technically a breaking change. 

However, IMO this is not consistent with how it behaves for the creation case, where the `names` are equal to `null`(https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/415/files#diff-ab7b846c8b4a5b9482d558f599ee7fe427d8f12d6b9f3b7f00b506cd78d76bf6R298). I tend to leave this change for the sake of consistency. But, please, let me know if should preserve the existing behavior.

tests/integration/save_test.js
ro0gr: @patocallaghan thanks for the suggestion. I've improved tests a bit.

tests/unit/fragment_array_test.js
```diff
@@ -282,4 +282,66 @@ module('unit - `MF.fragmentArray`', function(hooks) {
...
+        type: 'person',
+        id: 1,
+        attributes: {
+          names: null
+        }
+      }
+    });
+
+    run(() => {
+      assert.strictEqual(person.names, null);
```
ro0gr, line 298: this behavior stays unchanged

tests/unit/fragment_array_test.js
patocallaghan: Thanks @ro0gr, yep I agree that consistency is better and fragmentArrays can be nullable both when they are created and when they are updated. Technically we are still in the 5.x beta phase so a breaking change should be fine.

@knownasilya do you have thoughts on this slight change in behaviour?

knownasilya: Sounds fine to me, as long as we document the change.

patocallaghan:
It looks like [CI is acting up](https://github.com/adopted-ember-addons/ember-data-model-fragments/runs/4540746551?check_suite_focus=true). I'll try have a look today to see what's up

```
2/4] Fetching packages...
error An unexpected error occurred: "registry.yarnpkg.com/webidl-conversions/-/webidl-conversions-6.1.0.tgz: Request failed \"503 Service Unavailable\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/runner/work/ember-data-model-fragments/ember-data-model-fragments/yarn-error.log".
info Visit yarnpkg.com/en/docs/cli/install for documentation about this command.
Error!
1
undefined
Error: Process completed with exit code 1.
```

tests/unit/fragment_array_test.js
ro0gr: @knownasilya ok, I looked through the README and could not find a proper place where this detail could be injected into(in a natural way). I would gladly update the README(or whatever other documentation source) if there is a good idea on how to do it.

Meanwhile, I've updated the commit message, and PR description to contain a `[breaking]` label, + provided some more context regarding the reasons for this change. So I expect it to become a part of the changelog.
Also at this point we've fixed the intended behavior in tests, which could be approached as a kind of documentation as well. I think a `[breaking]` entry in the changelog + tests may be a good enough source of documentation. WDYT?

patocallaghan: Documenting it as `[breaking]` in the changelog sounds good enough to me 👍

patocallaghan:
I have a PR in #419 to see if we can clean make sure the `ember-release` try scenario doesn't fail

patocallaghan:
@ro0gr do you want to rebase against master as [I merged a fix](https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/419) so we can run CI against the different ember data versions?

ro0gr:
@patocallaghan thanks! Sure, will rebase in few hours.

patocallaghan:
Thanks @ro0gr!

knownasilya:
Do we want to do a beta release?

patocallaghan:
@knownasilya yep I think we should.

any thoughts on landing https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/384 first though? While this fix allows us to set fragments/arrays to null, there's another unrelated bug which prevents you from updating a fragment array to a non-null value afterwards.

ro0gr:
I'd also support a new beta.

patocallaghan:

ro0gr:
@patocallaghan Cool. Not sure if it's just a timing issue. But seems the version has not been published to NPM? Still shows 5.0.0-beta.2 as the latest.

patocallaghan:
@ro0gr ah apologies. I've pushed the new version to npm just now

Issue #406, opened 2021-08-25
Model fragments broken against Ember Data 3.28


kevinkucharczyk:
I was testing Ember Data Model Fragments against the latest Ember Data - version 3.28 - and unfortunately it looks like they're not compatible. Here's what happens when I try to save a model that contains a fragment:
![Screen Shot 2021-08-26 at 7 30 55 am](https://user-images.githubusercontent.com/3127419/130867504-5953b5b9-d2d3-471d-a96b-e663ad990b82.png)
```
index.js:178 Uncaught Error: Assertion Failed: You need to pass a model name to the store's serializerFor method
    at assert (index.js:178)
    at Store.serializerFor (ext.js:230)
    at Store.superWrapper [as serializerFor] (index.js:442)
    at Class.serializeFragment [as serialize] (fragment.js:40)
    at ApplicationSerializer.serializeAttribute (json.js:1105)
    at json.js:1035
    at -private.js:1611
    at Array.forEach (<anonymous>)
    at Snapshot.eachAttribute (-private.js:1610)
    at ApplicationSerializer.serialize (json.js:1034)
assert @ index.js:178
serializerFor @ ext.js:230
superWrapper @ index.js:442
serializeFragment @ fragment.js:40
serializeAttribute @ json.js:1105
(anonymous) @ json.js:1035
(anonymous) @ -private.js:1611
eachAttribute @ -private.js:1610
serialize @ json.js:1034
serialize @ rest.js:584
superWrapper @ index.js:442
serializeIntoHash @ rest.js:612
serializeIntoHash @ -private.js:634
createRecord @ rest.js:694
(anonymous) @ -private.js:1867
invokeCallback @ rsvp.js:493
(anonymous) @ rsvp.js:558
(anonymous) @ rsvp.js:19
invoke @ backburner.js:338
flush @ backburner.js:229
flush @ backburner.js:426
_end @ backburner.js:958
end @ backburner.js:708
_run @ backburner.js:1013
run @ backburner.js:752
run @ index.js:116
callback @ application.js:434
```

This particular problem seems to happen at https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/b1a661b3a8c07bd9d395769498062db8871d25f9/addon/transforms/fragment.js#L38
The `serialize` method is expecting a `Snapshot`, but actually receives the model instance of the fragment (I'm not sure why) - `modelName` is empty there.

Here's a minimal reproduction on a fresh Ember app:
[...]

PR #417, opened 2021-12-08
record-data implementation for ember-data 3.28+

dwickern:
An attempt at E-D 2.8 compatibility using record-data and public APIs.

Fragments are stored internally as RecordData:
```ts
type Fragment = RecordData | null;
type FragmentArray = RecordData[] | null;
type Array = unknown[] | null;

type FragmentData = Record<string, Fragment | FragmentArray | Array>;

interface FragmentRecordData {
  // fragment dirty state, analogous to _attributes
  _fragments: FragmentData; 

  // fragment in-flight state, analogous to _inFlightAttributes
  _inFlightFragments: FragmentData;

  // fragment canonical state, analogous to _data
  _fragmentData: FragmentData;
}
```

FragmentRecordData no longer references Fragment records. Fragment records are created on demand, either by calling `get` on the fragment attribute or `objectAt` on a fragment array.

FragmentArray and StatefulArray are based on ember-data's ManyArray.

/cc @runspired #406


addon/record-data.js
```diff
@@ -348,158 +475,366 @@ export default class FragmentRecordData extends RecordData {
       @private
     */
   changedAttributes() {
```
dwickern, line 477: `changedAttributes` previously returned `[ currentFragment, currentFragment ]` where `currentFragment` is a Fragment instance.

Now it returns `[ oldState, currentState ]` where `oldState` is the fragment's canonical state serialized and `currentState` is the fragment's dirty state serialized.

dwickern:
I couldn't figure out how to make snapshots work. Snapshots call `get` on the record, which returns the `Fragment` instance.
https://github.com/emberjs/data/blob/b16011382b2321171b18ab4d23eeea26ad1d7a7c/packages/store/addon/-private/system/snapshot.ts#L165

How can we make snapshots use the serialized fragment value instead?

knownasilya:
cc @runspired regarding that last comment, not sure if you have any ideas.

knownasilya:
@VincentMolinie I know you have contributed quite a bit, do you think this could be rebased against your changes? Or would it be better to open a new PR and pull in these changes onto master?

VincentMolinie:
@knownasilya yes of course, do you want me to do a new PR with theses changes ?

knownasilya:
@VincentMolinie that's up to you, what you think is easiest.

dwickern:
I'll see about merging master into this PR. It doesn't look too bad.

dwickern:
It will be tricky to support `typeKey` as a function for this use-case:

https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/bd9042d71f3ada648b1bc3a838db3e5b5847e80e/tests/dummy/app/models/component.js#L6-L9
https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/bd9042d71f3ada648b1bc3a838db3e5b5847e80e/tests/unit/serialize_test.js#L31-L43

The ComponentOptions typeKey depends on its owner's `type` attribute. Do we initialize the owner's attributes and _then_ its fragments so that we can pass a partially-constructed owner to `typeKey`? What if `typeKey` depends on the contents of some other fragment?

addon/record-data.js
```diff
@@ -2,369 +2,479 @@
...
+  isFragmentDirty(key) {
+    return this.__fragments?.[key] !== undefined;
+  }
+
+  getCanonicalState() {
+    const result = Object.assign({}, this._data);
+    for (const [key, behavior] of Object.entries(this._fragmentBehavior)) {
+      if (!this.hasFragment(key)) {
+        // force the fragment attribute to initialize its default value
+        internalModelFor(this).getRecord().get(key);
```
dwickern, line 459: This sucks. We can't serialize fragment state without involving the fragment attributes which are responsible for default values.

addon/record-data.js
```diff
@@ -375,170 +485,366 @@ export default class FragmentRecordData extends RecordData {
...
+    internalModelFor(this).adapterDidCommit(data);
+  }
+  _fragmentRollbackAttributes() {
+    internalModelFor(this).rollbackAttributes();
+  }
+  _fragmentCommitWasRejected() {
+    internalModelFor(this).adapterDidInvalidate();
+  }
+  _fragmentUnloadRecord() {
+    internalModelFor(this).unloadRecord();
```
dwickern, line 761: This all exists to keep the InternalModel state in sync with RecordData. Maybe that's not necessary any more?

addon/record-data.js
runspired: Yeah this is something I want to deprecate in ember-data too.

addon/record-data.js
runspired: Shouldn't be? There's no longer a state machine and IM has held no data directly for along time

dwickern: I've been testing against E-D 3.28.3. Are there more changes since then?

runspired: @dwickern the big flood comes in 4.5, basically the internals are fully rewritten at that point with another big rewrite to them in the works

addon/array/stateful.js
```diff
@@ -14,16 +14,22 @@ import { fragmentDidDirty, fragmentDidReset } from '../states';
...
     @property owner
     @type {DS.Model}
   */
-  owner: null,
+  get owner() {
+    const owner = this.recordData.getFragmentOwner();
+    if (!owner) {
+      return null;
+    }
+    return this.store._internalModelForResource(owner.identifier).getRecord();
```
runspired, line 31: you can just call `store.peekRecord(owner.identifier)`

addon/record-data.js
runspired: intenralModelFor helper will need adjusted for 4.5, that said I don't think you need it. You have access to the store and the identifier, just peek the record.

addon/record-data.js
runspired: as it should <3

addon/array/stateful.js
dwickern: Looks like that was added in 3.29 https://github.com/emberjs/data/commit/3bc81a6ee6ba980e7201c7bb936cee635e706274

runspired: @dwickern there is no 3.29, I think an alpha of 4.0 got tagged that way. Fairly sure we landed this for 3.28 let me check, but if not I can backport the capability to make it easier.

runspired: hrm yeah, you need 4.0. I can backport that to 3.28 because otherwise you don't have a way of retrieving newly created records that don't have an ID yet.

dwickern: I don't think we need it backported. We can use `_internalModelForResource` for 3.8 and `peekRecord` on 4+.

addon/record-data.js
dwickern: I tested a bit more against E-D 2.8 and 4.5. There's more than the InternalModel's state-- InternalModel is also responsible for notifying that the record attributes changed.

Take rollback for example. The InternalModel calls `InternalModel.notifyAttributes` with the return value from `RecordData.rollbackAttributes`. When we roll back nested fragments, we also need to fire notifications for those fragments' changed attributes.

richgt:
@dwickern - I'm running into issues on 3.28 now, and probably need this version of Fragments. What's the status of this PR? Is there anything I can do to help?

esbanarango:
> @dwickern - I'm running into issues on 3.28 now, and probably need this version of Fragments. What's the status of this PR? Is there anything I can do to help?

Same here! It would be great to have at least a beta version out of this PR 🙏

dwickern:
I got the tests passing against E-D 3.28. My merge missed a change from #432. However `typeKey`-as-a-function is not implemented by this PR and I'm not sure how it can be supported. We need to know the `typeKey` in order to construct the fragment, but we need to pass the fragment to the `typeKey` function??

Tests fail against older E-D versions because they need a [feature flag enabled](https://github.com/emberjs/data/blob/v3.16.9/packages/store/addon/-private/system/model/internal-model.ts#L1160). I'd rather drop support for older versions.

knownasilya:
@VincentMolinie  would we need to revisit that implementation?

runspired:
My guidance would be to get something that works 3.28->4.6

4.7+ changes are going to be a bit of a wade and the best course there is probably to just end support at 4.6 as ~4.11+ for data will come with this feature out of the box

dwickern:
To support E-D 4+ we'll need to upgrade ember-cli. I've done most of the work already but I'd like to keep it separate from this PR. Could we get this merged to v6-development branch?

knownasilya:
@patocallaghan what are your thoughts on that, should we make that breaking change for v5? v5 doesn't seem like it's ready for release without this PR...

dwickern:
v5 works well for me with ED 3.27. What about this support matrix:

MF v5 for ED versions [3.13, 3.28)
MF v6 for ED versions [3.28, 4.6)

patocallaghan:
@knownasilya @dwickern yep those versions sound reasonable 👍

deanmarano:
Upgrading from 3.24 -> 3.28 and this PR resolves an issue I'm seeing (reload not updating data). Is there help needed to get this merged and a beta released?

knownasilya:
Released v5.0.0-beta.9 today. If this looks good, I can release a v5 later this week if everyone is fine with that.

knownasilya:
Although looks like older versions on master are broken? Don't remember if this was the case before. There have been no new commits from me.

addon/record-data.js
```diff
@@ -375,170 +485,366 @@ export default class FragmentRecordData extends RecordData {
...
-        } else {
-          fragment._internalModel._recordData.commitWasRejected();
-        }
+      const inFlight = this._inFlightFragments[key];
+      if (inFlight != null) {
+        behavior.unload(inFlight);
       }
+      const fragmentData = this._fragmentData[key];
+      if (fragmentData != null) {
+        behavior.unload(fragmentData);
```
deanmarano, line 682: Testing this out against my test suite, I'm getting a bunch of errors from here.
```
Assertion Failed: You can only unload a record which is not inFlight.
at Object.assert (http://localhost:7357/app/assets/vendor.js:32841:15)
    at Object.assertAgainstUnloadRecord [as unloadRecord] (http://localhost:7357/app/assets/vendor.js:117846:12)
    at InternalModel.send (http://localhost:7357/app/assets/vendor.js:118720:32)
    at InternalModel.unloadRecord (http://localhost:7357/app/assets/vendor.js:118405:12)
    at FragmentRecordData._fragmentUnloadRecord (http://localhost:7357/app/assets/vendor.js:145645:30)
    at FragmentBehavior.unload (http://localhost:7357/app/assets/vendor.js:145096:16)
    at FragmentRecordData.unloadRecord (http://localhost:7357/app/assets/vendor.js:145565:20)
```

addon/array/fragment.js
```diff
@@ -81,15 +31,45 @@ const FragmentArray = StatefulArray.extend({
...
+    const fragment = this.store.createFragment(this.modelName, data);
+    setFragmentOwner(fragment, this.recordData, this.key);
+    return recordDataFor(fragment);
+  },
 
-    return normalizeFragmentArray(this, content, data, true);
+  replace(start, deleteCount, items) {
+    assert('The third argument to replace needs to be an array.', isArray(items));
+    const data = this.currentState.slice();
+    data.splice(start, deleteCount, ...items.map((item, i) => this._normalizeData(item, i)));
```
richgt, line 70:
```suggestion
    data.splice(start, deleteCount, ...items.map((item, i) => this._normalizeData(item, start + i)));
```

richgt: without this, if I have a fragment array [{name: Obj1}], and I try to addObject({name: Obj2}), I get two objects with {name: Obj2} because you use the index of the item in the added items array instead of the index of the item in the actual array to look for existing, therefore going back to [0] and modifying that in normalizeData.

addon/array/fragment.js
```diff
@@ -81,15 +31,45 @@ const FragmentArray = StatefulArray.extend({
...
+      `You can only add '${this.modelName}' fragments or object literals to this property`,
+      typeOf(data) === 'object' || isInstanceOfType(this.store.modelFor(this.modelName), data)
+    );
+
+    if (isFragment(data)) {
+      const recordData = recordDataFor(data);
+      setFragmentOwner(data, this.recordData, this.key);
+      return recordData;
+    }
+    const existing = this.objectAt(index);
```
richgt, line 57: This is the spot where sending index(0) in my suggested case will cause duplication. We don't want to actually modify the first object in the array, hence the `start + i` addition.

addon/array/fragment.js
richgt: Confirmed with a test - this only affects adding Objects, not full Fragments, since there's a return in there if the object you're adding is already a fragment. I have a PR with a test showing this. I'll create it.

richgt: https://github.com/dwickern/ember-data.model-fragments/pull/2

knownasilya:
v5 is out

addon/array/fragment.js
dwickern: Good find!

knownasilya:
@dwickern I wasn't aware this was done, since you kept adding commits 😄

addon/array/stateful.js
knownasilya: Was this resolved?

dwickern: I don't think it's blocking for 3.28, maybe something to address for ember-data v4

knownasilya (changes requested):
One small tweak :D Also could you post a basic changelog as a comment in this pr?

addon/ext.js
```diff
@@ -40,95 +35,16 @@ function serializerForFragment(owner, normalizedModelName) {
...
-    let identifier;
-    if (lte('ember-data', '3.13.0')) {
-      throw new Error('This version of Ember Data Model Fragments is incompatible with Ember Data Versions below 3.13. See matrix at https://github.com/lytics/ember-data-model-fragments#compatibility for details.');
-    }
-    if (gte('ember-data', '3.15.0')) {
-      identifier = this.identifierCache.getOrCreateRecordIdentifier({ type, id, lid });
-    } else {
-      identifier = { type, id, clientId: lid };
+    if (!gte('ember-data', '3.28.0')) {
+      throw new Error('This version of Ember Data Model Fragments is incompatible with Ember Data Versions below 3.28. See matrix at https://github.com/lytics/ember-data-model-fragments#compatibility for details.');
```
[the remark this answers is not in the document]
knownasilya: Or remove this section..

knownasilya:
@dwickern back to you

dwickern:
All breaking changes:

1. Requires ember-data 3.28

1. [changedAttributes](https://api.emberjs.com/ember-data/4.12/classes/Model/methods/changedAttributes?anchor=changedAttributes) now includes a deep copy of the fragment state in the form `[oldFragment, newFragment]`

1. `typeKey` as a function (#430) is not supported in this release

1. Fragments no longer fire unnecessary property change notifications when pushing data

    For example:
    
    ```js
    // models/person.js
    export default class Person extends Model {
     @fragment('name') name;
    
     @computed('name') // <-- missing dependent keys
     get fullName() {
       return `${this.name.first} ${this.name.last}`;
     }
    }
    
    // models/name.js
    export default class Name extends Fragment {
      @attr('string') first;
      @attr('string') last;
    }
    ```
    
    The `fullName` computed property incorrectly depends on `'name'`. In model-fragments 5.x, updating the `first` attribute via push would fire a change notification for `'first'` as expected but also for `'name'`. To fix this you should properly list the dependent keys. In this case change `@computed('name')` to `@computed('name.{first,last}')`.

    Note that this only affects legacy computed properties and only when pushing data via `store.push`, `store.pushPayload`, `model.save` response, etc. Calling `person.set('name.first', '...')` did not (and still does not) fire a property change notification for `'name'`.

    (credit to @richgt for finding this)

knownasilya (approved):

knownasilya:
v6.0.0 is out, supports 3.28 - 4.4. See the readme for compatibility details. Also https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/417#issuecomment-1546760562 ☝🏼

PR #419, opened 2021-12-18
Do not fail CI on Ember release

patocallaghan:
@knownasilya Any thoughts on this?

Until https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/406 is resolved CI is 🔴. Should we disable testing against Ember Data release until we land a fix there? It will remove noise on any PRs we land in the meantime. I think we should continue testing against ember-source release though

I've also added a note to the README about 3.28 not being fully supported yet

Edit: I've changed to opt out of ember-release altogether as lots of changes are required to make the test suite run in ember 4.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,12 +184,14 @@ jobs:
           ember-3.13,
           ember-lts-3.16,
           ember-lts-3.20,
-          ember-release,
+          ember-lts-3.24,
           ember-default-with-jquery,
           ember-classic
         ]
         allow-failure: [false]
         include:
+          - ember-try-scenario: ember-release
+            allow-failure: true
           - ember-try-scenario: ember-beta
             allow-failure: true
           - ember-try-scenario: ember-canary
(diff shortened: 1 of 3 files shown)


patocallaghan:
Ah we need `ember-auto-import@2` for the ember-release scenario

```
stack: Error: To use these addons, your app needs ember-auto-import >= 2: ember-source
```

knownasilya:
I'm fine with disabling for now

patocallaghan:
@knownasilya It appears a good chunk of work would need to happen to make the tests even compatible with ember-source 4.0, i.e. lots of dependencies and repo changes need to happen. I think to unblock we should just opt out of `ember-release` too for now. I can follow up with some work to modernize the repo.

knownasilya:
Agreed. Really that should be allowed to fail but I believe github actions don't support that yet?

patocallaghan:
It should be working now. I've opted ember-release into allowing to fail in the ci.yml file. Before it wouldn't even run the test suite

patocallaghan:
Or maybe I'm just wrong? 🤔 not sure where the 'allow-failure' syntax came from as it isn't listed in the GitHub Actions syntax docs 🤷🏻‍♂️

patocallaghan:
I'm going to merge this as-is

PR #428, opened 2022-02-23
fix: handle current state not yet set

VincentMolinie:
When doing the following:
```js
    const user = store.createRecord('user', {
      id: 1,
      firstName: 'Vince',
      lastName: 'Mol',
      email: 'firstLast@forestadmin.com',
      permissions: ['editProject'], // This is an `array` from fragment
    });
```
`ember-data-model-fragments` crash in that case because the `currentState` is not yet setup as the init of the record is not yet called but we try to access it in `fragmentDidDirty` and `fragmentDidReset`.

To fix it I first thought of just handling this case in the `fragmentDidDirty` and `fragmentDidReset`. But it actually makes no sense I think. So I added a new method `isStillInitializing` to prevent calling `fragmentDidDirty` and `fragmentDidReset` when the record is not yet initialized.

### DIFF
--- a/addon/array/stateful.js
+++ b/addon/array/stateful.js
@@ -179,6 +179,11 @@ const StatefulArray = ArrayProxy.extend(Copyable, {
     let record = get(this, 'owner');
     let key = get(this, 'name');
 
+    // Abort if fragment is still initializing
+    if (record._internalModel._recordData.isStillInitializing(key)) {
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


knownasilya:
Seems like there are more broken tests? At least when I compare against the last commit on master.

VincentMolinie:
Should be fixed. I added a check on ember-data version to handle both versions

PR #431, opened 2022-03-17
fix(dirty state): prevent record from being tagged as dirty when not changed

VincentMolinie:
This commit: https://github.com/adopted-ember-addons/ember-data-model-fragments/commit/f64a15ed57671923e7f630c9a5277541ea506bd4 introduced an issue on the dirty state. The attribute `_data` was always overridden even if the server does not send any data. 
This means that whenever you get a 204, the current state was `_data = {}` (an empty json). So whenever we try to update a property even to the same value the record is marked as dirty and the `changedAttributes` old value was always `undefined`

### DIFF
--- a/addon/ext.js
+++ b/addon/ext.js
@@ -85,14 +85,13 @@ assign(RecordDataPrototype, {
   },
 
   didCommit(data) {
-    // It seems the will commit has changed because _attributes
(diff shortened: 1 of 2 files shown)


knownasilya:
Released as v5.0.0-beta.5

PR #432, opened 2022-04-05
fix(serializer): fix the deserialization when using a typeKey as a function

VincentMolinie:
Deserialization for polymorphic attributes which have a typeKey as function doesn't work.
This PR address this bug


tests/dummy/app/models/person.js
```diff
@@ -10,6 +10,12 @@ export default Model.extend({
   titles: MF.array(),
   hobbies: MF.fragmentArray('hobby', { defaultValue: null }),
   houses: MF.fragmentArray('house'),
+  titleDescription: MF.fragment('title-description', {
+    polymorphic: true,
+    typeKey: (data, owner) => {
+      return `title-description-${owner.title}`;
+    }
+  }),
```
VincentMolinie, line 18: This produce the error

addon/attributes.js
```diff
@@ -20,6 +20,7 @@ function metaTypeFor(name, type, options) {
 
   if (options && options.polymorphic) {
     let typeKey = options.typeKey || 'type';
+    typeKey = typeof typeKey === 'function' ? '__dynamic__' : typeKey;
```
VincentMolinie, line 23: multiline function breaks the regex used on metaType. That's why we are replacing it it by a static value

addon/ext.js
```diff
@@ -331,6 +332,23 @@ JSONSerializer.reopen({
...
+    let attributes = get(typeClass, 'attributes');
+
+    typeClass.eachTransformedAttribute((key, typeClass) => {
+      if (data[key] === undefined) {
+        return;
+      }
+
+      let transform = this.transformFor(typeClass);
+      let transformMeta = attributes.get(key);
+      data[key] = transform.deserialize(data[key], transformMeta.options, data);
```
VincentMolinie, line 348: The only thing added compared to ember is the last argument which is the data of the parent

VincentMolinie:
@knownasilya could you please take a look 👼

iStefo:
Hi,

I'm just dropping in because I very recently started to upgrade/refactor an older ember app of mine. While integrating polymorphic fragment arrays, I noticed that deserialisation using `typeKey` functions was broken on 5.0.0-beta.5, but fixed by your patch. Or at least nearly fixed.

What I needed to adjust was to change the `FragmentArrayTransform` and pass `options` and `parentData` through, just as in `FragmentTransform`:

```js
  deserialize: function deserializeFragmentArray(data, options, parentData) {
    if (data == null) {
      return null;
    }

    return data.map(datum => {
      return this.deserializeSingle(datum, options, parentData);
    }, this);
  },
```

Now, I didn't want to PR your PR, but since you seem to be actively working on this, it should be easy to integrate this change for you. I'll report if I notice any other changes are required.

knownasilya:
@VincentMolinie I can merge once the above feedback was addressed I guess.

VincentMolinie:
@knownasilya  on it will be done in a few minutes 😉

VincentMolinie:
Done

VincentMolinie:
Could we have a release with that commit @knownasilya 🙏  ?

knownasilya:
Done

PR #437, opened 2022-05-12
fix(dirty state): fix dirtyness after save when using polymorphic fragment in some specific use case

VincentMolinie:
Some code was not updated. It seems `_fragments` has been replaced by `fragments` property but in some places we still use `_fragments`. The thing is when didCommit is called it use the `_fragments` to notify the nested attributes they were not dirty anymore


knownasilya (approved):
👍🏼

PR #438, opened 2022-06-07
fix(polymorphic): update of type of a polymorphic key works

VincentMolinie:
Updating a fragment to something that should be another fragment type doesn't work. I just make sure that when setting a fragment value the type is the same otherwise recreate the fragment

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -144,12 +144,14 @@ export default class FragmentRecordData extends RecordData {
         isInstanceOfType(store.modelFor(declaredModelName), value)
     );
 
+    // If the fragment was null before or if the new value i not of the same model as the previous one (polymorphism) we should recreate the fragment.
+    const shouldRecreateFragment = !fragment || (getActualFragmentType(declaredModelName, options, value, record) !== fragment.constructor.modelName);
     if (!value) {
       fragment = null;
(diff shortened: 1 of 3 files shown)


VincentMolinie:
@knownasilya needs your approval here if you got time 👼

addon/record-data.js
```diff
@@ -149,7 +150,7 @@ export default class FragmentRecordData extends RecordData {
     } else if (isFragment(value)) {
       // A fragment instance was given, so just replace the existing value
       fragment = setFragmentOwner(value, record, key);
-    } else if (!fragment) {
+    } else if (!fragment || (getActualFragmentType(declaredModelName, options, value, record) !== fragment.constructor.modelName)) {
```
knownasilya, line 153: I think we need a comment here, or maybe put this into a descriptive variable so we know what this is doing in the future.

knownasilya (changes requested):
Small change, otherwise it's 🚀

VincentMolinie:
Done @knownasilya thanks again for your quick reply 🙏

knownasilya (changes requested):
Tiny change

knownasilya (approved):

VincentMolinie:
Thank you so much @knownasilya 🙏

Issue #444, opened 2022-08-10
[Types] `FragmentArray` accepts registry key, should accept full fragment class


charlesfries:
The new unreleased type declarations want me to type `FragmentArray` like this:

```js
import type { FragmentArray } from 'ember-data-model-fragments/attributes';

interface Args {
  model: FragmentArray<'user'>;
}
```

and not like this:

```js
import type { FragmentArray } from 'ember-data-model-fragments/attributes';
import type UserModel from 'my-app/models/user';

interface Args {
  model: FragmentArray<UserModel>;
}
```

Shouldn't we be passing in the full class to the generic? Otherwise iterating over fragment array just yields the registry keys instead of the actual objects, according to the TypeScript declarations.

PR #443, opened 2022-08-01
chore: add typescript definitions files

VincentMolinie:

PR #445, opened 2022-08-11
chore: fix typescript declarations

charlesfries:
This PR fixes some type string vs type class inconsistencies in the TypeScript declaration files from #443 

#444

### DIFF
--- a/attributes.d.ts
+++ b/attributes.d.ts
@@ -1,35 +1,36 @@
 import Model from '@ember-data/model';
 import EmberArray from '@ember/array';
 import ComputedProperty from '@ember/object/computed';
+import Fragment from 'ember-data-model-fragments/fragment';
 import FragmentRegistry from 'ember-data-model-fragments/types/registries/fragment';
 import FragmentAttributesRegistry from 'ember-data-model-fragments/types/registries/fragment-attributes';
 // eslint-disable-next-line ember/use-ember-data-rfc-395-imports
 import TransformRegistry from 'ember-data/types/registries/transform';
 
-interface FragmentArray<FragmentType extends keyof FragmentRegistry> extends EmberArray<FragmentType> {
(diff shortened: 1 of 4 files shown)


attributes.d.ts
```diff
@@ -1,35 +1,36 @@
...
 import TransformRegistry from 'ember-data/types/registries/transform';
 
-interface FragmentArray<FragmentType extends keyof FragmentRegistry> extends EmberArray<FragmentType> {
-  addFragment(fragment: FragmentRegistry[FragmentType]): FragmentRegistry[FragmentType];
-  removeFragment(fragment: FragmentRegistry[FragmentType]): FragmentRegistry[FragmentType];
-  createFragment(attributes: FragmentAttributesRegistry[FragmentType]): FragmentRegistry[FragmentType];
+interface FragmentArray<T extends Fragment> extends EmberArray<T> {
+  addFragment(fragment: T): T;
+  removeFragment(fragment: T): T;
+  createFragment(attributes?: Record<string, unknown>): T;
```
charlesfries, line 13: This is basically how the normal `Store.createRecord` method is typed

knownasilya:
@charlesfries this ready for review?

charlesfries:
@knownasilya it is ready for review now. I was taking another look at `createFragment` and I realized why the original PR author possibly went with string keys rather than real fragment classes in the fragment generics--because afaik we can't start with the fragment class and get back to the string key, then over and up to the fragment attributes interface from its registry. if we start with just the key we can easily work our way to both registries

however I personally am not really in love with the idea of having to define a separate interface just for defining the raw structure of a fragment. I have seen `Partial<T>` used in other libs for this kind of problem, and so I have updated the PR to remove the attribute registry functionality in favor of `Partial`. lmk your thoughts

knownasilya (approved):

PR #464, opened 2023-05-16
Add support for ember-data 4.0 - 4.6

dwickern:
Now supports ember-data 3.28 to 4.6.

Ember-data 4.6 is as far as we can go for model-fragments v6. Ember-data 4.7 introduces RecordData v2 ([RFC 461](https://github.com/emberjs/rfcs/blob/master/text/0461-ember-data-singleton-record-data.md)) which will require significant changes.


addon/record-data.js
```diff
@@ -622,6 +621,7 @@ export default class FragmentRecordData extends RecordData {
 
   willCommit() {
     for (const [key, behavior] of Object.entries(this._fragmentBehavior)) {
+      this._initializeFragmentDefaultValue(key);
```
dwickern, line 624: ember-data's Snapshot constructor calls .get on every model attribute, which has the side effect of initializing fragments' default values. Ember-data v4 changed when it creates the snapshot.

In ember-data 3.28, hooks are called in this order:

    model.save -> snapshots created -> recordData.willCommit -> recordData.didCommit

And in ember-data 4+

    model.save -> recordData.willCommit -> snapshots created -> recordData.didCommit

To replicate the problem:
1. Construct a model containing a fragment with a default value
2. Immediately call model.save
3. RecordData.willCommit will ignore the fragment because it doesn't exist
4. ember-data's Snapshot will initialize the fragment in uncommitted state
5. RecordData.didCommit will fail because the fragment is not in-flight:

  > Error: Attempted to handle event `didCommit` on <name:null> while in state root.loaded.created.uncommitted. 

This line ensures that all default values are initialized by willCommit.

tests/unit/array_property_test.js
```diff
@@ -214,39 +215,41 @@ module('unit - `MF.array` property', function (hooks) {
...
 
-  test('supports array observers', async function (assert) {
-    store.push({
-      data: {
-        type: 'person',
-        id: 1,
-        attributes: {
-          nickName: 'Tyrion Lannister',
-          titles: ['Hand of the King'],
+  if (!gte('ember-data', '4.0.0')) {
```
dwickern, line 218: Technically, this should check the ember-source version. We could also remove the code from StatefulArray which handles array observers.

dwickern: Fixed in 8f98d19ff6791e32f72eee93e737347c0b3338b2

addon/record-data.js
dwickern: 4598753d0499d18f36aedde72889ff8a3c968fd7 makes RecordData responsible for default values instead of the attribute. We no longer have to invoke the attribute getter for its side effect and we no longer depend on the Snapshot constructor.

dwickern:
ready to review @knownasilya

knownasilya:
Should we release current master before merging this, or is it ok to merge this? Seems like it might be ok because it preserves 3.28 support? @dwickern

dwickern:
I think it's ok to merge this first, it has a few fixes for 3.28 also

package.json
```diff
@@ -58,12 +58,12 @@
...
     "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-terser": "^4.0.2",
-    "ember-data": "~3.28.13",
+    "ember-data": "~4.6.0",
     "ember-load-initializers": "^2.1.2",
     "ember-page-title": "^7.0.0",
     "ember-qunit": "^6.1.1",
     "ember-resolver": "^10.0.0",
-    "ember-source": "~3.28.12",
+    "ember-source": "~4.6.0",
```
knownasilya, line 66: @dwickern did you mean to keep these at 4.6?

knownasilya (approved):

package.json
dwickern: Yes, so that we're testing against 3.28 / 4.4 / 4.6

dwickern: Should we add ember-data as a peer dependency?

knownasilya: Not sure, is that recommended now @runspired ?

Issue #467, opened 2023-06-06
model.reload() don't update fragmentArray properties (using v6+ and ember-data 4.6)


ndekeister-us:
Hello!

First thanks a lot for the awesome work to make this repository compatible with Ember 4+ 🙏 

With version 6 of this addon and ember-data 4.6 we found a little issue, when using `model.reload()`, `fragmentArray` properties are not updated correctly

https://github.com/adopted-ember-addons/ember-data-model-fragments/assets/56396753/f18349d0-72d3-4e20-9520-1889b74784cf

If we are looking at `model.currentState.recordData.__fragmentData['myProperty']`, it contains correct values, but `model['myProperty']` still contains the initial value

Reproduced here -> https://github.com/ndekeister-us/ember-test-app_ember-data-model-fragments

deanmarano:
I've also experienced not seeing reload on ED 3.28. It looks like [this test for reloading](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/f4f4daf8901f8dd4a6fef398eb3b0cebae8cf3b4/tests/integration/save_test.js#L713) is skipped, and it is failing if ran.

deanmarano:
I checked out the sample repo here and [the PR that I got merged](https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/469) doesn't resolve this issue (because of [this rollback](https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/469/commits/45ad08c6ecd98ce3fbaf14eb27311e3cabad3659)). [This code](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/master/addon/record-data.js#L695-L697) is the issue (even though I thought it might not be necessary): 

```javascript
if ((updates[key] === null) !== (original[key] === null)) {
  changedKeys.push(key);
}
```
It handles if the fragment is null or is becoming null, but doesn't actually detect changes (as the `_changedFragmentKeys` function name implies).

I'm not sure what the best way to detect changes here. My initial attempt was checking if they `===` each other, which doesn't work for arrays (so it was always triggering, which _did_ fix the bug). I considered verifying on length and clientId:

```javascript
let wasOrBecameNull = original[key] === null || updates[key] === null;
let isDifferentLength = original[key]?.length !== updates[key].length;
let hasNewIds = !updates[key].every(record => original[key].includes(record.clientId));
if (wasOrBecameNull || isDifferentLength || hasNewIds) {
[...]

PR #469, opened 2023-06-23
Fix issue with reloading array records.

deanmarano:
I was looking at getting [reloading for fragment arrays working](https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/467), and I think I stumbled across a bug.

This conditional:
```
 if ((updates[key] === null) !== (original[key] === null))
```
it wasn't catching when the array existed and was different, so I added another condition `(updates[key] !== original[key])`. This updated the `_fragmentData` properly, but the cache was still being used so updates weren't making it through. I nabbed some code to clear the cache and the test started passing.

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -692,7 +692,10 @@ export default class FragmentRecordData extends RecordData {
       if (this._fragments[key]) {
         continue;
       }
-      if ((updates[key] === null) !== (original[key] === null)) {
+      if (
+        (updates[key] === null) !== (original[key] === null) ||
+        updates[key] !== original[key]
+      ) {
         changedKeys.push(key);
       }
     }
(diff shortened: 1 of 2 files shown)


addon/record-data.js
```diff
@@ -692,7 +692,10 @@ export default class FragmentRecordData extends RecordData {
       if (this._fragments[key]) {
         continue;
       }
-      if ((updates[key] === null) !== (original[key] === null)) {
+      if (
+        (updates[key] === null) !== (original[key] === null) ||
+        updates[key] !== original[key]
```
knownasilya, line 697: This is a very weird conditional, could this be split into some variables that describe what is going on?

knownasilya (changes requested):

addon/record-data.js
deanmarano: I agree about the lack of readability of this. I made a truth table to help me understand the original:

```
|                                               | updates[key]===null | original[key]===null | Result |
| --------------------------------------------- | ------------------- | -------------------- | ------ |
| { updates[key] = "12", original[key] = null } | FALSE               | TRUE                 | TRUE   |
| { updates[key] = "12", original[key] = "ab" } | FALSE               | FALSE                | FALSE  |
| { updates[key] = null, original[key] = null } | TRUE                | TRUE                 | FALSE  |
| { updates[key] = null, original[key] = "ab" } | TRUE                | FALSE                | TRUE   |
```

It made me think that the second case (where the values are different) should return `true`, but after playing around with it I didn't need to make a change, just call `notify`.

PR #472, opened 2023-07-07
Ensure changing arrays are notified.

### LINKED ISSUE #467: model.reload() don't update fragmentArray properties (using v6+ and ember-data 4.6)

deanmarano:
I identified [this code as problematic](https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/467#issuecomment-1624297238) previously, but the tests passed without it so [a partial fix was merged](https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/469). More tests are needed! This would fix #467.

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -1,5 +1,6 @@
 // eslint-disable-next-line ember/use-ember-data-rfc-395-imports
 import { RecordData } from 'ember-data/-private';
+import { diffArray } from '@ember-data/model/-private';
 import { recordDataFor } from '@ember-data/store/-private';
 import { assert } from '@ember/debug';
 import { typeOf } from '@ember/utils';
@@ -692,7 +693,11 @@ export default class FragmentRecordData extends RecordData {
       if (this._fragments[key]) {
         continue;
       }
-      if ((updates[key] === null) !== (original[key] === null)) {
+      const eitherIsNull = original[key] === null || updates[key] === null;
+      if (
+        eitherIsNull ||
+        diffArray(original[key], updates[key]).firstChangeIndex !== null
+      ) {
         changedKeys.push(key);
       }
     }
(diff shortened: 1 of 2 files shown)


addon/record-data.js
```diff
@@ -692,7 +692,14 @@ export default class FragmentRecordData extends RecordData {
       if (this._fragments[key]) {
         continue;
       }
-      if ((updates[key] === null) !== (original[key] === null)) {
+      const wasOrBecomingNull = original[key] === null || updates[key] === null;
+      let isChangedArray = original[key]?.length || updates[key]?.length;
+      if (isChangedArray) {
+        isChangedArray = !updates[key]?.filter?.(
+          (record) => !original[key]?.includes(record.clientId)
+        );
```
knownasilya, line 700: Could the updates have an entry for a new value? In that case this wouldn't be marked as changed, right?

knownasilya (changes requested):

addon/record-data.js
deanmarano: This first attempt has some issues. Looking at this a bit more, I see we use `diffArray` in `addons/array/stateful.js`. Maybe we should use that here as well?

knownasilya: Unfamiliar with it, but all for reuse.

deanmarano: Used that and added a test as well. Tested it with the example in #467 and it worked there as well.

knownasilya (approved):
❤️

Issue #470, opened 2023-06-28
Fragments with a function-style defaultValue error in v6


grayt0r:
Versions:
* `ember-data-model-fragments`: `v6.0.1`
* `ember-data`: `3.28.13`

When pushing a payload to the store for a model that contains a fragment with a function-style `defaultValue` I receive the following error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at -private.js:2170:1
    at changeProperties (index.js:589:1)
    at ExampleModel._notifyProperties (-private.js:2168:1)
    at InternalModel.setupData (-private.js:5991:1)
    at Store._load (-private.js:9153:1)
    at Store._pushInternalModel (-private.js:9380:1)
    at -private.js:9331:1
    at Backburner._run (backburner.js:843:1)
    at Backburner._join (backburner.js:822:1)
    at Backburner.join (backburner.js:632:1)
```

The model is defined like this:

```js
export default class ExampleModel extends Model {
  @attr name;

  @fragment('example-fragment', { defaultValue: () => ({}) }) exampleFragment;
}
```

The error goes away and everything works when I downgrade to `ember-data-model-fragments` `v5.0.0` and `ember-data` `v3.27.1`.

I can also fix the error by swapping to a non-function `defaultValue`, i.e.:

```js
export default class ExampleModel extends Model {
  @attr name;

  @fragment('example-fragment', { defaultValue: {} }) exampleFragment;
}
```

Is this a known limitation of `v6`? Or a bug?

A minimal reproduction can be found here: https://github.com/grayt0r/ember-data-model-fragments-bug ([relevant commit](https://github.com/grayt0r/ember-data-model-fragments-bug/commit/ccfc8fb5608a9a09cd323add8856f322368b2c6b)).

Thanks!

VincentMolinie:
Oh I actually got the exact same issue ! This PR might fix your issue (https://github.com/adopted-ember-addons/ember-data-model-fragments/pull/474)

grayt0r:
@VincentMolinie looks like it does... thank you 🙌

knownasilya:
Released as v6.0.2

PR #474, opened 2023-07-13
fix(polymorphism): fix regression introduced with ember-data-model-fragments 6.0.0 where owner is not passed to typeKey function

VincentMolinie:
Will also:
- fix adopted-ember-addons/ember-data-model-fragments#470

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -603,7 +603,8 @@ export default class FragmentRecordData extends RecordData {
     const type = getActualFragmentType(
       definition.modelName,
       definition.options,
-      attributes
+      attributes,
+      this._fragmentGetRecord()
     );
     const recordData = this.storeWrapper.recordDataFor(type);
     recordData.setFragmentOwner(this, definition.name);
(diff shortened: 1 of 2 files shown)


addon/record-data.js
```diff
@@ -715,12 +717,24 @@ export default class FragmentRecordData extends RecordData {
         // strip fragments from the attributes so the super call does not process them
         delete data.attributes[key];
 
+        subFragmentsToProcess.push({ key, behavior, canonical, attributes });
+      }
+    }
+
+    // Wee need first the attributes to be setupd before the fragment to be able to access them (for polymorphic fragments for example)
```
VincentMolinie, line 724:
```suggestion
    // Wee need first the attributes to be setup before the fragment, to be able to access them (for polymorphic fragments for example)
```

addon/record-data.js
```diff
@@ -732,13 +746,12 @@ export default class FragmentRecordData extends RecordData {
...
     }
 
-    const changedAttributeKeys = super.pushData(data, calculateChange);
     const changedKeys = mergeArrays(changedAttributeKeys, changedFragmentKeys);
     if (gte('ember-data', '4.5.0') && changedKeys?.length > 0) {
       internalModelFor(this).notifyAttributes(changedKeys);
     }
     // on ember-data 2.8 - 4.4, InternalModel.setupData will notify
-    return changedKeys;
+    return changedKeys || [];
```
VincentMolinie, line 754: This line fix this issue https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/470

knownasilya: Should this be a `??`?

VincentMolinie: This methods should always return an array as it is actually doing a `changedKeys.length` afterward. So I prefer to be safe

knownasilya (approved):

PR #476, opened 2023-07-19
changedAttributes still return the dirty attributes until the record has been fully committed

VincentMolinie:
When trying to upgrade to the latest version of ember-data-model-fragments I realised there was an issue inside the `changedAttributes()`. 
`changedAttributes()` is supposed to return all the dirty attributes until the record has been fully committed not just when calling the `save` method

In the test I show what we do, we use it to send only the dirty attributes

### DIFF
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -619,7 +619,7 @@ export default class FragmentRecordData extends RecordData {
 
   hasChangedFragments() {
     return (
-      this.__fragments !== null && Object.keys(this.__fragments).length > 0
+      this._fragmentsOrInFlight !== null && Object.keys(this._fragmentsOrInFlight).length > 0
     );
   }
 
@@ -669,7 +669,7 @@ export default class FragmentRecordData extends RecordData {
 
   changedFragments() {
     const diffData = Object.create(null);
-    for (const [key, newFragment] of Object.entries(this._fragments)) {
+    for (const [key, newFragment] of Object.entries(this._fragmentsOrInFlight || {})) {
       const behavior = this._fragmentBehavior[key];
       const oldFragment =
         key in this._inFlightFragments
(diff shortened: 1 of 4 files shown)


tests/dummy/app/serializers/component.js
```diff
@@ -1,3 +1,21 @@
...
+    if (data.data?.attributes) {
+        // NOTICE: Remove all the unchanged attributes in the payload.
+        const changedAttributes = Object.keys(record.changedAttributes());
+
+        Object.entries(data.data.attributes).forEach(([attributeName, value]) => {
+          if (!changedAttributes.includes(attributeName)) {
+            delete data.data.attributes[attributeName];
+          }
+        });
+    }
```
knownasilya, line 17: What is this for?

tests/unit/serialize_test.js
```diff
@@ -319,4 +320,67 @@ module('unit - Serialization', function (hooks) {
...
+          }
+        });
+        return [204, { 'Content-Type': 'application/json' }, '{}'];
+      });
+
+      await component.save();
+
+      assert.deepEqual(component.changedAttributes(), {});
+    })
+  });
```
knownasilya, line 385: Nice!

knownasilya:
Looks like there's a linting issue

addon/record-data.js
```diff
@@ -669,7 +669,7 @@ export default class FragmentRecordData extends RecordData {
 
   changedFragments() {
     const diffData = Object.create(null);
-    for (const [key, newFragment] of Object.entries(this._fragments)) {
+    for (const [key, newFragment] of Object.entries(this._fragmentsOrInFlight || {})) {
```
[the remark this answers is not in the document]
VincentMolinie: true 👍

tests/dummy/app/serializers/component.js
VincentMolinie: It's written in the comment above `// NOTICE: Remove all the unchanged attributes in the payload.`

knownasilya: Why is this required for the test?

VincentMolinie: It's what the test is about, checking that the changedAttributes is correct between the `save` and the return of the `save` (So between `willCommit` and `didCommit`)

knownasilya (approved):

knownasilya:
Released as v6.0.3

PR #479, opened 2023-09-01
Add `copy` to type declarations

charlesfries:
- [Defined here](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/11e263b11c911a23c430db8bd2b77a1f40eef68c/addon/fragment.js#L91)


knownasilya (approved):

knownasilya:
Released

PR #480, opened 2023-09-23
fix(changedAttributes): initial value of property which are supposed to be null return a value when the record is in flight

VincentMolinie:


knownasilya:
Shouldn't the test fail in all scenarios?

tests/unit/fragment_test.js
```diff
@@ -201,6 +201,51 @@ module('unit - `MF.Fragment`', function (hooks) {
...
+      null,
+      'old fragment is indicated in the diff object'
+    );
+    assert.deepEqual(
+      newName,
+      { first: 'Rob', last: 'Stark', prefixes: [] },
+      'new fragment is indicated in the diff object'
+    );
+
+    person._internalModel.adapterWillCommit();
```
VincentMolinie, line 233: I don't know if there is a better way to pass the record in a willCommit state

VincentMolinie:
@knownasilya yeah it was not ready sorry 😉

VincentMolinie:
@knownasilya it's ready 😉

VincentMolinie:
Can anyone look at it please 👼  ?

knownasilya:
I'll have a look tomorrow, or Monday.

knownasilya (approved):
Nice

knownasilya:
Released

Issue #427, opened 2022-02-14
Auto tracking regression in +3.28


pjcarly:
I have narrowed down a regression in our application to Ember Data Model Fragments, we notice this since upgrading to 3.28 in certain scenarios.

I have tried debugging, but can't seem to figure out where the regression occurs.
It happens in a context the moment we try to call `model.get('address')`. Where address is a fragment.
We often see this as a warning, but in some specific scenario's it causes an outright error, crashing the application.

We are seeing this error:

```
index.js:129 Uncaught (in promise) Error: Assertion Failed: You attempted to update `_tracking` on `Tag`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

```

Seeing something similar here: https://github.com/emberjs/ember.js/issues/18613#issuecomment-1029901855
And thanks to the explanation of @NullVoxPopuli I kind of understand what is going on. Unfortunately, the solution presented `await Promise.resolve();` does not fix the issue.

I will try to create a reproduction application, in the coming days.

patocallaghan:
Thanks @pjcarly. When you say 3.28 do you mean ember-source or ember-data? But yeah a reproduction of some sort would be good 👍

pjcarly:
Both at 3.28, I have been trying to reproduce it the entire week, no luck so far. Still investigating.

pjcarly:
By using `msw.js` instead of `ember-cli-mirage` I was able to reproduce the issue in this repository:

https://github.com/pjcarly/autotracking-regression-repro

ptgamr:
Any updates on this ? We're hitting the same issue on our way to 4...

knownasilya:
Have you tried the betas?

ptgamr:
@knownasilya No I haven't, will try and let you know... Thank you!

knownasilya:
@ptgamr did you have luck?

averydev:
@knownasilya @ptgamr Any luck here? I'd love to get up to 4!

knownasilya:
I'll be releasing a new patch version today or tomorrow, so watch out for that as well.

VincentMolinie:
I'm on `ember-data-model-fragment@6.0.5` and I'm getting the same issue 🤔
It seems to happen whenever I do a `createRecord` on the store which contains a fragment / fragment array

VincentMolinie:
[...]

PR #482, opened 2023-12-12
fix: prevent You attempted to update _tracking on Tag issue on Ember-data 4.5+

VincentMolinie:
I realised we had an issue on, that I reproduce pretty easily in my app with Ember-data 4.6. 
The RecordState is registered when we create a new Model.
For a classic native model, the RecordState will be build after all the initial value of the attributes have been set.
But with fragments it's the opposite, the RecordState is build first then the attributes are set which result in a duplicate update of the state

This produce the exact same error that you can find on this issue: https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/427

My fix is based on the RecordDataDefault, I saw they are using `notifyPropertyChange` to notify an attribute has change instead of what we are doing which is using notifyStateChange to say the state of the record has changed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,7 +178,7 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          [ember-lts-3.28, ember-lts-4.4, ember-classic]
+          [ember-lts-3.28, ember-lts-4.4, ember-4.6, ember-4.8, ember-classic]
         allow-failure: [false]
         include:
           - ember-try-scenario: ember-release
--- a/addon/record-data.js
+++ b/addon/record-data.js
@@ -967,12 +967,21 @@ export default class FragmentRecordData extends RecordData {
   }
 
   notifyStateChange(key) {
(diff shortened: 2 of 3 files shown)


VincentMolinie:
The PR is not ready yet, I'll check why the CI is not apssing

VincentMolinie:
Okey it should be good now

knownasilya:
Looks like the change is for 4.5+, but you mentioned 4.6, was that a typo?

VincentMolinie:
> Looks like the change is for 4.5+, but you mentioned 4.6, was that a typo?

Yes correct sorry. Because I thought at first it was just 4.6 but you are right it happens also for 4.5.0

knownasilya (approved):

knownasilya:
Does this solve #427

VincentMolinie:
> Does this solve #427

I don't think so as it's related to Ember 3.28. But I actually had the issue on 3.28 but before `ember-data-model-fragments@6.0.5`. So this issue might already be fixed anyway

knownasilya:
Version released

PR #483, opened 2024-02-02
Fix #316 polymorphic create fragment

dwickern:
I fixed the merge conflicts from #436 and got it working with the new RecordData implementation.

Fixes #316
Closes #436

PR #486, opened 2024-03-13
fix passing arbitrary props to `fragmentArray.createFragment`

dwickern:
When calling `createRecord` or `createFragment`, it's possible to pass arbitrary extra props which are not defined on the model. #483 broke this behavior when calling `fragmentArray.createFragment(props)`


knownasilya (approved):

knownasilya:
@dwickern 6.0.8 released

PR #487, opened 2024-05-03
Fix updating fragment arrays multiple times in the same runloop

dwickern:
In production builds, certain fragment array mutations produce the wrong result. For example, this code swaps the first two items in the array:
```js
const first = fragmentArray.objectAt(0)
fragmentArray.removeAt(0)
fragmentArray.insertAt(1, first)
```

In development, this works as expected. But in a production build, this code results in the first item being duplicated (as if the `removeAt` did not happen).

There are already some failing tests for this issue, as long as we run the tests in production mode:
```
not ok 19 Chrome 124.0 - [23 ms] - integration - Dependent State: reordering a fragment array dirties the fragment array and owner record
    ---
        actual: >
            3
        expected: >
            2
        stack: >
                at Object.<anonymous> (http://localhost:7357/assets/tests-28c38f66d4715374e954e49f4d4e8a30.js:39:22)
        message: >
            fragment array length is maintained
        negative: >
            false
        browser log: |
    ...

not ok 110 Chrome 124.0 - [23 ms] - unit - `MF.fragmentArray`: changes to array contents change the fragment array `hasDirtyAttributes` property
    ---
        actual: >
            false
        expected: >
            true
        stack: >
                at Object.<anonymous> (http://localhost:7357/assets/tests-28c38f66d4715374e954e49f4d4e8a30.js:306:193)
        message: >
            fragment array is returned to clean state
        negative: >
            false
        browser log: |
    ...
```

Some of the existing tests depend on `Ember.assert` throwing an exception. To make this work in production mode, I've replaced `assert.throws` with `expectAssertion`, which is similar to [ember-qunit-assert-helpers](https://github.com/workmanw/ember-qunit-assert-helpers) and to [ember's own expectAssertion helper](https://github.com/emberjs/ember.js/blob/v4.12.4/packages/internal-test-helpers/lib/ember-dev/assertion.ts).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: "14.x"
+  NODE_VERSION: "18.x"
 
 jobs:
   lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: "18.x"
+  NODE_VERSION: "14.x"
 
 jobs:
   lint:
@@ -77,7 +77,7 @@ jobs:
             ${{ runner.os }}-${{ matrix.node-version }}-
 
       - name: Install Dependencies
-        run: yarn install --frozen-lockfile
+        run: yarn install --frozen-lockfile --ignore-engines
         if: |
           steps.cache-dependencies.outputs.cache-hit != 'true'
 
@@ -121,7 +121,7 @@ jobs:
             ${{ runner.os }}-${{ matrix.node-version }}-
 
       - name: Install Dependencies
-        run: yarn install --frozen-lockfile
+        run: yarn install --frozen-lockfile --ignore-engines
         if: |
           steps.cache-dependencies.outputs.cache-hit != 'true'
 
@@ -163,7 +163,7 @@ jobs:
             ${{ runner.os }}-${{ matrix.node-version }}-
 
       - name: Install Dependencies
-        run: yarn install --no-lockfile --non-interactive
+        run: yarn install --no-lockfile --non-interactive --ignore-engines
 
       - name: Test
         run: yarn test:ember --launch ${{ matrix.browser }}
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: "14.x"
+  NODE_VERSION: "18.x"
```
knownasilya, line 47: Is this required for this change? If we aren't upgrading the version, I'd prefer we keep it at the lowest supported so we don't miss issues.

dwickern: I had to upgrade ember-try after hitting https://github.com/ember-cli/ember-try/issues/967
> Run node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO --- ember test --environment=production
> Error!
> Error: Cannot copy '../../../browserslist/cli.js' to a subdirectory of itself, '../../../browserslist/cli.js'.
> Error: Cannot copy '../../../browserslist/cli.js' to a subdirectory of itself, '../../../browserslist/cli.js'.
>     at /home/runner/work/ember-data-model-fragments/ember-data-model-fragments/node_modules/ember-try/node_modules/fs-extra/lib/copy/copy.js:210:21
>     at FSReqCallback.oncomplete (fs.js:180:23)

And then ember-try v3 has a higher node requirement:
> error ember-try@3.0.0: The engine "node" is incompatible with this module. Expected version "16.* || >= 18.*". Got "14.21.3"

Node 16 was EOL in Sep 2023 so I went with v18.

If we want to stick with node 14, `yarn --ignore-engines` might work.

knownasilya: I'm fine with upgrading to 18, we'd just need to update the readme and any other references, and do a major bump.

dwickern: `--ignore-engines` seems to work in CI. ember-try is a devDependency so it shouldn't affect the consuming app. I was able to install locally with node 14 and yarn 1.22 without needing to use `--ignore-engines`.

knownasilya (approved):

knownasilya:
Released as v6.0.9

PR #507, opened 2026-02-02
Support ember-data 4.13.x and require extending FragmentStore/FragmentSerializer

RobbieTheWagner:
This PR adds support for Ember Data 4.13 and updates the addon docs and fragment integration APIs around explicit FragmentStore, FragmentSerializer, and related setup.

This is a breaking change because consumers now need to extend the fragment-aware store and serializer classes in their apps instead of relying on the older implicit store patching.

A follow-up PR will stack on top of this branch to add Ember Data 5.8 support separately.


RobbieTheWagner:
@coderabbitai review

PR #514, opened 2026-02-26
Guard fragment access on destroyed/unloaded records

deanmarano:
Prevent crashes and infinite recursion when fragment properties are accessed on destroyed or unloaded owner records.

- Guard Fragment.toString() and toStringExtension() against destroyed fragments to break the infinite recursion cycle where recordIdentifierFor -> assertion -> String(record) -> toString() loops
- Guard fragment, fragmentArray, and array attribute getters to return null when the owner record is destroying/destroyed
- Remove dependent keys from fragmentOwner() CP and add a destroyed guard, preventing Ember's 'Attempted to access computed on a destroyed object' assertion during teardown
- Add tests verifying toString() returns a safe string on destroyed fragments without crashing

### DIFF
--- a/addon/attributes/array.js
+++ b/addon/attributes/array.js
@@ -55,6 +55,9 @@ export default function array(type, options) {
   // eslint-disable-next-line ember/require-computed-property-dependencies
   return computed({
     get(key) {
+      if (this.isDestroying || this.isDestroyed) {
+        return null;
+      }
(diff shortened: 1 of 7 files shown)


MelSumner (approved):

RobbieTheWagner (approved):

RobbieTheWagner:
Thanks for the PR @deanmarano!

PR #516, opened 2026-03-26
Add warp-drive/ember-data 5.8 compatibility

deanmarano:
## Summary

This adds compatibility with warp-drive/ember-data 5.8 while keeping existing 4.12 support working.

The 5.8 release introduced some fairly significant internal changes -- the Snapshot class moved to `@ember-data/legacy-compat/-private`, Store and Serializer no longer support `.reopen()`, and the schema service filters out non-standard attribute kinds (which breaks our fragment/fragment-array/array kinds). This PR works through all of those issues with a mix of conditional imports, prototype patching, and runtime schema service patching.

Key changes:
- Import `Snapshot` from `@ember-data/legacy-compat/-private` instead of the old location
- Patch `Model.attributes` to include fragment-kind computed properties that 5.8 now filters out
- Patch the schema service's `attributesDefinitionFor` at runtime to re-include fragment attributes
- Add `@warp-drive/build-config` setup in `ember-cli-build.js` (with graceful fallback for older versions)
- Configure `FragmentCache` to optionally wrap a pre-existing inner cache (needed because 5.8's Store creates its own `JSONAPICache` before our `createCache` override runs)
- Add `skipNotify` support in `FragmentStateManager` to avoid autotracking errors during record creation
- Handle fragments in a disconnected state during `teardownRecord`

I think the approach is pretty solid but there might be edge cases I haven't hit yet. The test suite passes for both 4.12 and 5.8, which feels like a good sign.

## Test plan

- All existing tests pass against ember-data 4.12
- All existing tests pass against ember-data/warp-drive 5.8
- CI matrix includes both `ember-data-4.12` and `ember-data-5.8` scenarios

## Release Notes

* **Chores**
  * Upgraded support to Ember Data 5.8 and related ecosystem packages
  * Refactored internal fragment state management and cache architecture to align with newer Ember Data versions
  * Enhanced continuous integration testing with expanded scenario matrix for broader compatibility verification
  * Updated build configuration and development dependencies for modern tooling compatibility
  * Improved serializer and schema handling for enhanced fragment operations


RobbieTheWagner:
Thanks for all the great work here! There were a few things I tweaked a bit and did differently and now the latest release should support 4.12 through 5.x, so we should not need this PR.

PR #534, opened 2026-05-07
Convert to a v2 addon

aklkv:


.github/workflows/ci.yml
```diff
@@ -12,80 +12,82 @@ concurrency:
...
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v6
-      - uses: pnpm/action-setup@v5
-        with:
-          version: 10
-      - name: Install Node
-        uses: actions/setup-node@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
mansona: I understand that this is a security thing where you pin the actions to a sha, but we don't usually do that in this org and it's too much of a hassle for us to maintain at scale 😞 

can you swap it back to v6 please (and the rest of them to their respective versions) 🙏

blueprints/fragment/index.js
mansona, line 1: is there a reason we remove the blueprints? 🤔

lib/calculate-version.cjs
mansona, line 1: is this file even used any more? it looks like it was used in the lib/version broccoli shenanigans 🤔 but we can't broccoli any more in v2 😂

blueprints/fragment/index.js
aklkv: can we keep them? are they working in v2 format?

lib/calculate-version.cjs
aklkv: I replicated functionality with rollup but I was not sure if we even need this anymore

blueprints/fragment/index.js
mansona: yes blueprints work from v2 addons, I don't know about **these** blueprints but the idea of blueprints work 😂

config/ember-cli-update.json
```diff
@@ -0,0 +1,17 @@
+{
+  "schemaVersion": "1.0.0",
+  "projectName": "ember-data-model-fragments",
+  "packages": [
+    {
+      "name": "@embroider/addon-blueprint",
```
RobbieTheWagner: When converting to v2, isn't this just the normal ember blueprint now?

src/index.js
```diff
@@ -20,7 +24,6 @@ import FragmentSerializer, {
   @main ember-data-model-fragments
 */
 const MF = Namespace.create({
-  VERSION: VERSION,
```
RobbieTheWagner, line 23: So I am not sure what we were using `VERSION` for exactly, but maybe we should import from package.json?

config/ember-cli-update.json
aklkv: good catch, updated

src/index.js
aklkv: added this version which now derives directly from package.json

RobbieTheWagner:
Pushed 0573bfe, which replaces the hand-written `declarations/*.d.ts` with declarations built from source, since the declarations folder of a v2 addon should be generated:

- **All of `src/` is now TypeScript** (~5k lines converted). Public API surfaces carry real types — `fragment()`/`fragmentArray()`/`array()`/`fragmentOwner()`, `Fragment`, `FragmentStore#createFragment`, `FragmentArray`/`StatefulArray`, transforms, and serializers — including the `FragmentRegistry`-driven generics the manual declarations introduced. The registry augmentation point is preserved at the same module path (`ember-data-model-fragments/types/registries/fragment`). Internals that reach into ember-data private API are typed pragmatically with `any` (the unsafe-`any` lint family is disabled for `src/**/*.ts` with a comment explaining why).
- **`addon.declarations()` is enabled** in `rollup.config.mjs` (it was scaffolded but commented out), so `pnpm build` / `prepack` now emits `declarations/` via `ember-tsc --declaration --project tsconfig.publish.json`. The checked-in `declarations/` files are deleted and the built output is gitignored.
- **Two missing dev deps added**: `@ember/library-tsconfig` (which `tsconfig.publish.json` already extended but wasn't installed — the declaration build had never been runnable) and `@babel/preset-typescript` for both babel configs.
- **ember-data 5.x `unstable-preview-types` are mapped via tsconfig `paths`** in both tsconfigs — ember-data ships no `exports`-wired types, so without this every `@ember-data/*` import was implicitly `any`/erroring in .ts files.
- Classic `.extend()` classes (`Fragment`, the transforms, `StatefulArray`/`FragmentArray`) get explicit `declare class`/interface type surfaces with a cast, because the inferred `.extend()` types can't be emitted in declarations (TS4023 on ember's private symbols).

One deliberate typing divergence from the manual declarations: `FragmentArray#addFragment`/`removeFragment` are typed as returning the array (`this`), matching what `addObject`/`removeObject` actually return at runtime — the old `.d.ts` said they return the fragment, which was inaccurate.

Verified locally: `pnpm lint` (format, hbs, js, types, publish — publint passes against the packed tarball), `pnpm build` emits a full `declarations/` tree with docblocks intact, and `pnpm test` passes 234/235 (1 pre-existing skip).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

RobbieTheWagner (approved):

